PR #55, opened 2020-11-25
feature: Always accept identity observable factors

speller26:
This enables simultaneous measurement of qubit-wise commuting Pauli
words.

Also added validation to make sure Hermitian observables on more than 1
qubit that share a target qubit share all target qubits and in the same
order.

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/simulator.py
+++ b/src/braket/default_simulator/simulator.py
@@ -198,12 +198,13 @@ class DefaultSimulator(BraketSimulator):
         qubit_observable_mapping = {}
         identity_qubits = set()
         for result_type in observable_result_types:
+            observable = result_type.observable
             obs_obj = (
-                DefaultSimulator._tensor_product_index_dict(result_type.observable, lambda x: x)
-                if isinstance(result_type.observable, TensorProduct)
-                else result_type.observable
+                DefaultSimulator._tensor_product_index_dict(observable, lambda x: x)
+                if isinstance(observable, TensorProduct)
+                else observable
             )
-            measured_qubits = result_type.observable.measured_qubits
+            measured_qubits = observable.measured_qubits
             if measured_qubits is None:
                 measured_qubits = list(range(qubit_count))
 
@@ -212,11 +213,11 @@ class DefaultSimulator(BraketSimulator):
                     f"Result type ({result_type.__class__.__name__}) Observable "
                     f"({obs_obj.__class__.__name__}) references invalid qubits {measured_qubits}"
                 )
-            new_obs = DefaultSimulator._observable_hash(result_type.observable)
+            hashed_observable = DefaultSimulator._observable_hash(observable)
             for i in range(len(measured_qubits)):
                 DefaultSimulator._assign_observable(
                     obs_obj,
-                    new_obs,
+                    hashed_observable,
                     measured_qubits,
                     i,
                     not_none_observable_list,
@@ -231,7 +232,7 @@ class DefaultSimulator(BraketSimulator):
     @staticmethod
     def _assign_observable(
         whole_observable,
-        hashed_qubit_observable,
+        hashed_observable,
         measured_qubits,
         target_index,
         not_none_observable_list,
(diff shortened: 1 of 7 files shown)


src/braket/default_simulator/simulator.py
```diff
@@ -202,32 +212,96 @@ def _validate_and_consolidate_observable_result_types(
...
+                f"Qubit {qubit} is already being measured in {cls_existing};"
+                f" cannot measure in {cls_new}."
+            )
+        if cls_existing == Hermitian.__name__:
+            if not np.allclose(existing.matrix, new.matrix):
+                raise ValueError(
+                    f"Qubit {qubit} is already being measured in {existing.matrix};"
+                    f" cannot measure in {new.matrix}."
+                )
+            qubits_existing = existing.measured_qubits
```
licedric, line 289: Can you explain what this chunk of code is supposed to do? In particular, are `existing` and `new` supposed to be single-qubit or multi-qubit observables?

speller26: `existing` and `new` can be either single or multi-qubit _non-tensor product_ obeservables (so the latter case only applies to arbitrary Hermitian observables). The purpose of this method is to ensure that for any qubit that already has an observable measurement, and subsequent non-identity observables are the same (qubit-wise commuting arbitrary Hermitians are not yet supported).

The final branch, which is only reached for multi-qubit observables, ensures that the target qubits don't conflict. For example, let's say we have a 2-qubit Hermitian acting on [1, 2]. If we add the same Hermitian to qubit 1, we want to make sure that 1 is the first target qubit, and the second target is 2, basically ruling out targets of, say, [2, 1] or [0, 1].

src/braket/default_simulator/simulator.py
```diff
@@ -202,32 +212,96 @@ def _validate_and_consolidate_observable_result_types(
...
+            qubits_new = new.measured_qubits
+            if (
+                qubits_existing is not None
```
licedric, line 292: Are the first two checks (`qubits_existing is not None and len(qubits_existing) > 1`) necessary? Is it possible that `len(qubits_existing) == 1` but `len(qubits_new) > 1`?

src/braket/default_simulator/simulator.py
licedric: Thanks for the explanation.

src/braket/default_simulator/simulator.py
```diff
@@ -202,32 +212,96 @@ def _validate_and_consolidate_observable_result_types(
...
+                if isinstance(whole_observable, dict)
+                else whole_observable
+            )
+            if not existing_observable:
+                identity_qubits.discard(qubit)
+                qubit_observable_mapping[qubit] = qubit_observable
+
+                if (
+                    not none_observable_mapping.get(hashed_qubit_observable)
+                    and qubit_observable.measured_qubits.index(qubit) == 0
```
licedric, line 266: I spent a while being confused on this check, so a comment would probably help here.

src/braket/default_simulator/simulator.py
```diff
@@ -202,32 +212,96 @@ def _validate_and_consolidate_observable_result_types(
...
+            if qubit not in qubit_observable_mapping:
+                identity_qubits.add(qubit)
+            # Do nothing if non-identity observable already exists on the qubit
+        else:
+            qubit_observable = (
+                whole_observable[target_index]
+                if isinstance(whole_observable, dict)
+                else whole_observable
+            )
+            if not existing_observable:
```
licedric, line 260: What if existing_observable is the identity?

src/braket/default_simulator/simulator.py
speller26: The `if not np.allclose(existing.matrix, new.matrix)` check above ensures that `len(qubits_existing) == len(qubits_new)` if neither `qubits_existing` and `qubits_new` are `None` , so `len(qubits_existing) == 1 and len(qubits_new) > 1` is impossible. If `qubits_existing is None`, then `existing` must be a single-qubit observable so don't have to proceed with further checks. At this point, I suppose I can remove `len(qubits_existing) > 1`, since `len(qubits_existing) == 1 and qubits_existing != qubits_new` is impossible.

licedric: Let's remove `len(qubits_existing) > 1`. It confuses me a bit when the conditions aren't symmetric.

src/braket/default_simulator/simulator.py
speller26: Impossible. The check `if hashed_qubit_observable == Identity.__name__:` above ensures that identities will not have been added at this point. I'll add a comment about that.

licedric (approved):

PR #71, opened 2021-06-01
feature: get unitary for an entire circuit

indisoluble:
*Issue #, if available:* https://github.com/aws/amazon-braket-sdk-python/issues/209

*Description of changes:* New simulator to produce the unitary matrix that represents an entire circuit

*Testing done:* Enough to cover 100% of the code, however more might be required, please let me know

**Notice:** These changes are pointing to another PR https://github.com/aws/amazon-braket-schemas-python/pull/92 where I added a new result named `UnitaryMatrix`. This PR has to be merged before this one and, when it is done, I will update this one to point again to AWS `main` branch

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/unitary_matrix_simulation.py
+++ b/src/braket/default_simulator/unitary_matrix_simulation.py
@@ -1,0 +1,112 @@
+# Copyright 2019-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+from string import ascii_lowercase, ascii_uppercase
+from typing import List, Tuple
+
+import numpy as np
+import opt_einsum
+
+from braket.default_simulator.operation import GateOperation
+from braket.default_simulator.simulation import Simulation
+
+
+class UnitaryMatrixSimulation(Simulation):
+    """
+    This class tracks the evolution of the unitary matrix of a quantum system with
+    `qubit_count` qubits. The unitary of system evolves by applications of `GateOperation`s
+    using the `evolve()` method.
+    """
+
+    def __init__(self, qubit_count: int, shots: int):
+        """
+        Args:
+            qubit_count (int): The number of qubits being simulated.
+            shots (int): The number of samples to take from the simulation.
+                0 is the only valid value for this simulator.
+        """
+        super().__init__(qubit_count=qubit_count, shots=shots)
+        initial_unitary = np.eye(2 ** qubit_count, dtype=complex)
+        self._unitary_matrix = initial_unitary
+
+    def evolve(self, operations: List[GateOperation]) -> None:
+        """Evolves the unitary of the simulation under the action of
+        the specified gate operations.
+
+        Args:
+            operations (List[GateOperation]): Gate operations to apply for
+                evolving the unitary of the simulation.
+
+        Note:
+            This method mutates the state of the simulation.
+        """
+        self._unitary_matrix = UnitaryMatrixSimulation._apply_operations(
+            self._unitary_matrix, self._qubit_count, operations
+        )
+
+    @staticmethod
+    def _apply_operations(
+        unitary: np.ndarray, qubit_count: int, operations: List[GateOperation]
+    ) -> np.ndarray:
+        un_tensor = np.reshape(unitary, qubit_count * [2, 2])
+        for operation in operations:
+            matrix = operation.matrix
+            targets = operation.targets
+            un_tensor = UnitaryMatrixSimulation._apply_operation(
+                un_tensor, qubit_count, matrix, targets
+            )
+
+        return np.reshape(un_tensor, 2 * [2 ** qubit_count])
+
+    @staticmethod
+    def _apply_operation(
+        un_tensor: np.ndarray, qubit_count: int, matrix: np.ndarray, targets: Tuple[int, ...]
+    ) -> np.ndarray:
+        subscripts = UnitaryMatrixSimulation._einsum_subscripts(targets, qubit_count)
+
+        target_count = len(targets)
+        gate_matrix = np.reshape(matrix, target_count * [2, 2])
+
+        return opt_einsum.contract(subscripts, gate_matrix, un_tensor, dtype=complex, casting="no")
+
+    @staticmethod
+    def _einsum_subscripts(targets: Tuple[int, ...], qubit_count: int) -> str:
+        target_count = len(targets)
+        assert target_count + qubit_count <= len(ascii_lowercase), (
+            "Simulation qubit count + gate target count "
+            f"can not be bigger than {len(ascii_lowercase)}"
+        )
+
+        un_left_labels = ascii_lowercase[:qubit_count]
+        un_right_labels = ascii_uppercase[:qubit_count]
+
+        gate_left_labels = ascii_lowercase[: -1 - target_count : -1]
+        gate_right_labels = "".join(un_left_labels[-1 - target] for target in reversed(targets))
+
+        result_left_labels = list(un_left_labels)
+        for pos, target in enumerate(targets):
+            result_left_labels[-1 - target] = gate_left_labels[-1 - pos]
+        result_left_labels = "".join(result_left_labels)
+
+        return (
+            f"{gate_left_labels}{gate_right_labels}, "
+            f"{un_left_labels}{un_right_labels} -> "
+            f"{result_left_labels}{un_right_labels}"
+        )
+
+    @property
+    def unitary_matrix(self) -> np.ndarray:
+        """
+        np.ndarray: The unitary matrix equivalent to the entire simulation.
+        """
+        return self._unitary_matrix
--- a/test/unit_tests/braket/default_simulator/test_unitary_matrix_simulation.py
+++ b/test/unit_tests/braket/default_simulator/test_unitary_matrix_simulation.py
@@ -1,0 +1,65 @@
+# Copyright 2019-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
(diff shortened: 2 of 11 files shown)


src/braket/default_simulator/unitary_matrix_simulation.py
```diff
@@ -0,0 +1,112 @@
...
+    ) -> np.ndarray:
+        subscripts = UnitaryMatrixSimulation._einsum_subscripts(targets, qubit_count)
+
+        target_count = len(targets)
+        gate_matrix = np.reshape(matrix, target_count * [2, 2])
+
+        return opt_einsum.contract(subscripts, gate_matrix, un_tensor, dtype=complex, casting="no")
+
+    @staticmethod
+    def _einsum_subscripts(targets: Tuple[int, ...], qubit_count: int) -> str:
```
indisoluble, line 83: This code is based in this one in Qiskit Terra: https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/providers/basicaer/basicaertools.py#L132

test/unit_tests/braket/default_simulator/test_unitary_matrix_simulation.py
```diff
@@ -0,0 +1,65 @@
...
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+import numpy as np
+import pytest
+
+from braket.default_simulator import gate_operations
+from braket.default_simulator.unitary_matrix_simulation import UnitaryMatrixSimulation
+
+evolve_testdata = [
```
indisoluble, line 20: I wrote these tests based on the tests for the unitary simulator in Qiskit Terra: https://github.com/Qiskit/qiskit-terra/blob/main/test/python/basicaer/test_unitary_simulator.py

shiyunon (commented):
Several questions:

1. I don't see the code throw any error messages if the circuit has components that have nounitary matrix definitions. These components include: 1. quantum noises 2. (in the future) compiler options like `barrier` or classical operations 3. (in the future) parametric gates. Were these corner cases considered?
2. If you implemented a unitary simulation, why encode UnitaryMatrix as a Result type but not just let users do `circ.unitary()`?
3. It's not clear to me if add another "simulator" is necessary if we already have LocalSimulator and SV1.

shiyunon (commented):
Thanks for the PR. 

I have several questions before a full review from the team:

1. I don't see the code throw any error messages if the circuit has components that don't have unitary matrix definitions. These components include: 1. quantum noises 2.  compiler options like `barrier` or classical operations (this is future-proof, Braket doesn't have it yet) 3. parametric gates  (this is also future-proof, Braket doesn't have it yet).
2. If you implemented a unitary simulation, why encode UnitaryMatrix as a result type but not just let users do `circ.unitary()`?
3. It's not clear to me if add another "simulator" is necessary if we already have LocalSimulator and SV1.

indisoluble:
Hi @shiyunon, thanks for your comments :) Let's go one by one:

1. In order to implement `UnitaryMatrixSimulation.evolve()`, I followed the example of `StateVectorSimulation.evolve()` and `DensityMatrixSimulation.evolve()`; in both cases, although I might be wrong about this, they get the `matrix` & `targets` from a `GateOperation` instance and then perform the necessary operations. As I see it, it is on  `GateOperation` to validate the information returned but this might be a wrong assumption, if there is an example somewhere else in the code that I can follow, please let me know and thanks on advance :+1: 
2. Well, given that I coded a new simulator, I do need a new result type, otherwise [BaseLocalSimulator._validate_ir_results_compatibility()](https://github.com/indisoluble/amazon-braket-default-simulator-python/blob/circuit_unitary/src/braket/default_simulator/simulator.py#L120) will always fail, right?... although your next question is: why did I code a new simulator? Anyway, if by `circ` you mean the `braket.circuits.Circuit`, I have not modified [amazon-braket-sdk-python](https://github.com/aws/amazon-braket-sdk-python) ... yet :)
3. `braket.devices.LocalSimulator` is part of [amazon-braket-sdk-python](https://github.com/aws/amazon-braket-sdk-python). I have modified [amazon-braket-default-simulator-python](https://github.com/aws/amazon-braket-default-simulator-python) to provide a new device that `braket.devices.LocalSimulator` can use to extend its capabilities.

Regards.

shiyunon:
> Hi @shiyunon, thanks for your comments :) Let's go one by one:
> 
> 1. In order to implement `UnitaryMatrixSimulation.evolve()`, I followed the example of `StateVectorSimulation.evolve()` and `DensityMatrixSimulation.evolve()`; in both cases, although I might be wrong about this, they get the `matrix` & `targets` from a `GateOperation` instance and then perform the necessary operations. As I see it, it is on  `GateOperation` to validate the information returned but this might be a wrong assumption, if there is an example somewhere else in the code that I can follow, please let me know and thanks on advance 👍
> 2. Well, given that I coded a new simulator, I do need a new result type, otherwise [BaseLocalSimulator._validate_ir_results_compatibility()](https://github.com/indisoluble/amazon-braket-default-simulator-python/blob/circuit_unitary/src/braket/default_simulator/simulator.py#L120) will always fail, right?... although your next question is: why did I code a new simulator? Anyway, if by `circ` you mean the `braket.circuits.Circuit`, I have not modified [amazon-braket-sdk-python](https://github.com/aws/amazon-braket-sdk-python) ... yet :)
> 3. `braket.devices.LocalSimulator` is part of [amazon-braket-sdk-python](https://github.com/aws/amazon-braket-sdk-python). I have modified [amazon-braket-default-simulator-python](https://github.com/aws/amazon-braket-default-simulator-python) to provide a new device that `braket.devices.LocalSimulator` can use to extend its capabilities.
> 
> Regards.

1. Got it, I saw your supported gate list. Yes, it is indeed the case, simulators just validate circuits like QPUs, so there shouldn't be problems with it.
2 &3. In that case, I think it will be a better user experience if we can have this in the braket_sdk (at least for small circuits), but these design problems will probably have to be answered after the team's full review.

Thanks again for the PR!

indisoluble:
By the way, I checked the "dependent tests" that actually failed (like this one: https://github.com/aws/amazon-braket-default-simulator-python/pull/71/checks?check_run_id=2723103546#step:5:2973) and it seems like it is not able to find `braket.ir.jaqcd.UnitaryMatrix` but I did this change ... in my fork for [amazon-braket-schemas-python](https://github.com/indisoluble/amazon-braket-schemas-python). I guess the test got the `main` branch for AWS `amazon-braket-schemas-python` instead of mine. Hopefully, tests will pass once (and if) https://github.com/aws/amazon-braket-schemas-python/pull/92 is merged.

src/braket/default_simulator/unitary_matrix_simulation.py
nonhermitian: If using Qiskit code here, it would be nice to acknowledge that fact.

indisoluble: Sure :) Although this code is not a copy of that one in Qiskit so, as I see it, it does not feel under the requirements of Apache license ... but I have to admit I am not sure I fully understand all the details of Apache license, I might be wrong about this. Nonetheless, it is worth to reference the sources, should I add a comment in the code? Some advice on this regard would be appreciated.

test/unit_tests/braket/default_simulator/test_unitary_matrix_simulation.py
nonhermitian: This should have a copyright statement of Qiskit somewhere in the file as this is obvious a derivative work of the original.

src/braket/default_simulator/unitary_matrix_simulation.py
nonhermitian: So, per Apache 2, "derivative works" must contain the copyright as well.  However there is of course gray area.  Perhaps here just cite the original source of the code in question.  However, it is really up to the maintainers of this repo to decide such things.

indisoluble: Thank you @nonhermitian for your explanation!!! I will wait then for a comment from the reviewers. Regards

kshitijc:
Thank you for the PR @indisoluble! We respect the open source policies and would like to have attribution notices where necessary. 
Please allow some time for us to review this request, as this is an important feature and requires some collaborative decisions on the team.

kshitijc:
Thanks for your patience @indisoluble! After a careful review, we think that the unitary matrix representation functionality would be better suited to live in the amazon-braket-sdk-python repository, and be accessible to customers through a method on the Circuit class (`as_unitary()`) allowing the users to get the unitary representation of the circuit easily.  Here are some considerations which guided us towards this decision:
* Placing this functionality in a simulator would involve much more work, both from the users (the unitary representation of a Circuit now needs to be accessed in a roundabout way through a simulator) and from the maintainers of this code (there is additional code that goes with a device, which needs to be maintained).
* This requires a UnitaryMatrix result type, which can be a source of confusion for some of the customers (who are directly interacting with the service) since no other devices support it.
* There’s performance overhead in translating from the circuit class to the representation expected by the simulator, which can be avoided.

shiyunon:
Hi, @indisoluble. I complement @kshitijc 's review with the following points:

As for the Qiskit code that was used, in general, we have no problem with reusing code from another framework as long as we abide by the requirements of the Apache licence. However, it indeed requires more considerations for some of the choices here. On line 80 in `unitary_matrix_simulation.py`, the `contract` method from the`opt_einsum` package was used. We find it might be worthwhile to compare betewen the `opt_einsum` package and default Numpy `einsum`. Here are the tradeoffs:
1. `opt_einsum` is much faster for tensor contractions with many contract paths to choose from. For this simple usage, we expect the performance difference to be minimal for this use case but it requires further benchmarking.
2.  Using`opt_einsum` adds 3rd party dependence to Braket which increases complexity in future maintenance.
 
With the above said, it is beyond the scope of this PR to make the decision. We will respect your choice here and will further investigate into it on the team (so you don't have to do anything). The changes we do recommend you to make are:
1. For the `_einsum_subscripts` defined on line 83 in `unitary_matrix_simulation.py`, we recommend to implement this method using the `np.array` calling style instead of the `str` calling style to improve the readability. Specifically, both `einsum` and `opt_einsum` allow you to call `einsum` (or `contract` in this case) in two ways: (1) supply the subscripts as strings (like what you did here) (2) supply the subscripts as numpy.array. An example would be:
```
>>> a
array([[1, 2],
       [3, 4]])
>>> opt_einsum.contract('ij->ji', a)
array([[1, 3],
       [2, 4]])
>>> opt_einsum.contract(a, [0, 1], [1, 0])
array([[1, 3],
       [2, 4]])
```
We recommend the second way of supplying the subscripts to increase readability.

2. For line 85-85, in `unitary_matrix_simulation.py`, with the `numpy.array` way of supplying the subscripts, there shouldn't be a limit of 26 for the indices anymore. However, for user visibility, we do want the function to warn the users that `as_unitary()` can get slow if the qubit count is beyond 10 qubits. Please add the proper warning to this function.
3. Please add more test cases in test file `test_unitary_matrix_simulation.py`  to have a better test coverage. For example, the current test doesn’t test all gates. 

If you would make the above changes, we will review the new commit to see if an acknowledgement of Qiskit is needed. 

Again, thanks for your contribution here to Braket!

indisoluble:
Hi @kshitijc & @shiyunon,

Thanks for your comments, I will apply them (hopefully soon) in a new PR for  `amazon-braket-sdk-python`. Right now, I will close this PR as it is no longer necessary.

Regards.

PR #73, opened 2021-08-09
Calculate exact statistics from trace

speller26:
*Issue #, if available:*

*Description of changes:*

Currently, expectation and variance are calculated for `shots=0` by diagonalizing the state in the (mutually commuting) observables' shared basis and then multiplying by eigenvalues. This commit switches the implementation over to computing the inner product/trace directly for each observable.

*Testing done:*

## Merge Checklist

#### General

#### Tests


licedric:
Can you add a summary to this PR?

src/braket/default_simulator/observables.py
```diff
@@ -303,6 +331,15 @@ def measured_qubits(self) -> Tuple[int, ...]:
...
         return self._eigenvalues
 
+    def apply(self, state: np.ndarray) -> np.ndarray:
+        final = np.array(state)
+        for factor in self._factors:
+            final = factor.apply(final)
+        return final
+
+    def apply_to_qubit(self, state: np.ndarray, qubit: int) -> np.ndarray:
+        raise TypeError("Tensor product cannot by measured on each qubit individually")
```
licedric, line 341:
```suggestion
        raise TypeError("Tensor product cannot be measured on each qubit individually")
```

src/braket/default_simulator/density_matrix_simulation.py
```diff
@@ -149,15 +151,19 @@ def state_with_observables(self) -> np.ndarray:
             raise RuntimeError("No observables applied")
         return self._post_observables
 
+    def expectation(self, with_observables: np.ndarray) -> float:
```
licedric, line 154: Needs comment

src/braket/default_simulator/linalg_utils.py
```diff
@@ -0,0 +1,109 @@
...
+        targets (List[int]): The qubits of the output reduced density matrix;
+            if no target qubits are supplied, this method returns the trace of the density matrix.
+
+    Returns:
+        np.ndarray: The partial trace of the density matrix.
+    """
+    target_set = set(targets) if targets else set()
+    nkeep = 2 ** len(target_set)
+    idx1 = [i for i in range(qubit_count)]
+    idx2 = [qubit_count + i if i in target_set else i for i in list(range(qubit_count))]
```
licedric, line 106:
```suggestion
    idx2 = [qubit_count + i if i in target_set else i for i in range(qubit_count)]
```

speller26: Why the extra list instantiation? The range object achieves the goal

src/braket/default_simulator/result_types.py
```diff
@@ -294,10 +281,10 @@ def __init__(self, observable: Observable):
...
 
     @staticmethod
-    def _calculate_from_prob_distribution(
-        probabilities: np.ndarray, eigenvalues: np.ndarray
-    ) -> float:
-        return (probabilities @ eigenvalues).real
+    def _calculate_from_application(
+        simulation: Simulation, apply_func: Callable[[np.ndarray], np.ndarray]
+    ):
+        return simulation.expectation(apply_func(simulation.state_as_tensor))
```
licedric, line 287: I don't understand how this works after looking at `expectation`. Can you elaborate?

src/braket/default_simulator/linalg_utils.py
licedric: Exactly, hence my suggestion :)

speller26: Oh right, suggestion comes after

src/braket/default_simulator/state_vector_simulation.py
```diff
@@ -155,6 +157,11 @@ def state_with_observables(self) -> np.ndarray:
             raise RuntimeError("No observables applied")
         return self._post_observables
 
+    def expectation(self, with_observables: np.ndarray) -> float:
```
licedric, line 160: Needs comment too

src/braket/default_simulator/density_matrix_simulation.py
speller26: This overrides `expectation` defined in `Simulation`, and that's where the docs live.

src/braket/default_simulator/state_vector_simulation.py
speller26: ditto

src/braket/default_simulator/result_types.py
speller26: Implements abstract method in `ObservableResultType`; see docstring there

src/braket/default_simulator/state_vector_simulation.py
licedric: For that matter, why does this function take `with_observables` as an input instead of the observable itself? It feels a little bit confusing to me.

speller26: That's what I wanted to do, but the implementation gets complicated because variance requires two applications of the observable rather than one, but the method uses the internal state of the `Simulation` object. For example, if `expectation is defined as follows:
```python
def expectation(O):
    return ⟨ψ | O | ψ⟩
```

However, variance needs `O`^2,  and operator algebra has not been implemented. Alternatively, I could either add a `num_applications` field to `.apply`, which gives the further benefit of allowing precomputation of the power, or have `expectation` take in both the observable and some state, but both of these are relatively clunky.

That's why I elected to just let `expectation` take O|ψ⟩ instead of O.

licedric: I think the current implementation is clunky too - not only does it need psi but it also needs O |psi>, and there is some repeated information being passed around here.

It seems cleaner to me to pass `O`; and either have a separate variance function, or pass `O` as a callable function (so to get `O^2` you just pass the function applied twice).

speller26: The callable way is actually my favorite by far; the only reason I didn't do it was that I was worried about ease of use. If you have no problems with it, I'll happily make that change.

speller26: Working it out a bit more, it looks like even the callable approach is annoying; I'm just going to implement `__pow__` and open the algebra floodgates.

licedric: Why's the callable approach annoying? (And it's no less usable than what you have now I think?)

speller26: No less usable that what I currently have, but if I'm going to pass in a callable, I might as well optimize it so that observable powers are taken before applying. As well, I can exploit the involutority of the Paulis and Hadamard to speed up the calculation.

src/braket/default_simulator/observables.py
```diff
@@ -52,11 +57,42 @@ def measured_qubits(self):
...
+
+    Note: this class does not enforce that the matrix is Hermitian or involutory.
+    """
+
+    @abstractmethod
+    def __init__(self, matrix: np.ndarray, targets: Optional[List[int]] = None):
+        self._matrix = matrix
+        self._targets = _validate_and_clone_single_qubit_target(targets)
+
+    def __pow__(self, power, modulo=None):
```
licedric, line 82: This needs documentation. What happens when `power` is not an integer?
Also, what's `modulo`?

src/braket/default_simulator/result_types.py
```diff
@@ -94,56 +93,39 @@ def observable(self):
...
+    def calculate(self, simulation: Simulation) -> Union[float, List[float]]:
+        """Calculates the result type using the underlying observable.
 
-    @staticmethod
-    @abstractmethod
-    def _calculate_from_prob_distribution(
-        probabilities: np.ndarray, eigenvalues: np.ndarray
-    ) -> float:
-        """Calculates a result from the probabilities of eigenvalues.
+        Returns a a real number if the observable has defined targets,
```
licedric, line 99:
```suggestion
        Returns a real number if the observable has defined targets,
```

src/braket/default_simulator/state_vector_simulation.py
```diff
@@ -155,6 +149,13 @@ def state_with_observables(self) -> np.ndarray:
             raise RuntimeError("No observables applied")
         return self._post_observables
 
+    def expectation(self, observable: Observable) -> float:
+        qubit_count = self._qubit_count
+        with_observables = observable.apply(np.reshape(self._state_vector, [2] * qubit_count))
```
licedric, line 154: Great! Much more readable, and opens the door to future optimization if we want to special case `apply` for specific observables.

src/braket/default_simulator/observables.py
speller26: This is a [standard library function](https://docs.python.org/3/library/operator.html#operator.__pow__), and we don't [docstring these](https://github.com/aws/amazon-braket-sdk-python/blob/main/src/braket/circuits/observable.py#L73) in any of our libraries. However, I made a bit of a workaround.

`modulo` got included by the IDE for some reason, so I thought it was part of the signature; I'll remove it.

licedric:
What's the code path when computing expectations for `shots > 0`?

speller26:
> What's the code path when computing expectations for `shots > 0`?

Handled entirely by the SDK; result types won't be visible to this simulator, and it'll instead return samples.

src/braket/default_simulator/observables.py
```diff
@@ -52,11 +57,42 @@ def measured_qubits(self):
...
+        return Identity([qubit])
+
     def diagonalizing_gates(self, num_qubits: Optional[int] = None) -> Tuple[GateOperation, ...]:
         return ()
 
 
-class Hadamard(Observable):
+class _InvolutoryMatrixObservable(Observable, ABC):
+    """
+    An observable defined by an involutory matrix.
```
licedric, line 72: Give a better self-contained description (i.e. the matrix squared is the identity).

licedric (approved):

PR #75, opened 2021-08-20
feature: Calculate arbitrary observables

speller26:
Allows arbitrary observables to be attached to a circuit, allowing, for example, the simultaneous calculation of all terms of a Hamiltonian.

This change only applies for `shots=0`, as result types are only defined for `shots=0` in this simulator. The `shots>0` case is handled entirely by the SDK, using samples generated by the simulator.

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/simulator.py
+++ b/src/braket/default_simulator/simulator.py
@@ -13,9 +13,8 @@
 
 import uuid
 import warnings
-from typing import Any, Callable, Dict, List, Tuple, Union
+from typing import Any, Callable, Dict, List, Union
 
-import numpy as np
 from braket.device_schema.device_action_properties import DeviceActionType
 from braket.device_schema.simulators import GateModelSimulatorDeviceCapabilities
 from braket.ir.jaqcd import Program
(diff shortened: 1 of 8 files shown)


licedric:
> Allows arbitrary observables to be attached to a circuit, allowing, for example, the simultaneous calculation of all terms of a qubit Hamiltonian when shots=0.

Clarify this a bit more - it should be clear what happens. Namely, the user can create a circuit with arbitrary observables, but an error is returned if it is run with shots > 0.

speller26:
`shots=0` is actually only relevant for the SDK here; in the default simulator, result types are only defined for `shots=0`, and the `shots>0` case is handled entirely by the SDK, using samples generated by the simulator.

src/braket/default_simulator/simulator.py
```diff
@@ -379,26 +247,16 @@ def _observable_hash(observable: Observable) -> Union[str, Dict[int, str]]:
...
-
-        results = [0] * len(circuit_ir.results)
-
-        for index in non_observable_result_types:
-            results[index] = ResultTypeValue.construct(
+        return [
+            ResultTypeValue.construct(
                 type=circuit_ir.results[index],
-                value=non_observable_result_types[index].calculate(simulation),
+                value=result_types[index].calculate(simulation),
```
licedric, line 256: Any idea why we separated `non_observable` from `observable` previously?

licedric:
> shots=0 is actually only relevant for the SDK here; in the default simulator, result types are only defined for shots=0, and the shots>0 case is handled entirely by the SDK, using samples generated by the simulator.

Add this comment to the description then, since I think this may be a commonly asked question.

src/braket/default_simulator/simulator.py
speller26: The observable result types were extracted and their observables validated to ensure they were identical. Since that validation is no longer necessary, they don't have to be separated anymore. Honestly, they didn't even have to be split up so early even then, as the entire list could've just been filtered for the validation; doesn't matter now.

licedric (approved):

PR #93, opened 2022-07-21
Local sim branch

ajberdy:
*Issue #, if available:*

*Description of changes:*

Merge OpenQASM and Jaqcd simulators back into the main simulator class chain

Follow up action item of including OpenQASM language features in device capabilities will be implemented in a separate PR.

*Testing done:*

Added testing for branching and restructured existing correctness testing for both IRs.

Note: a lot of the existing correctness testing is implemented with Jaqcd. For now, that is ok because both OQ3/Jaqcd simulators use the same underlying simulation. The correctness for the OpenQASM part of the implementation is tested extensively separately from the Jaqcd tests. Additionally, this functionality will also be doubly tested in the BDK, where all Jaqcd correctness tests will also be converted to OpenQASM tests and run against the local simulator.

## Merge Checklist

#### General

#### Tests


src/braket/default_simulator/density_matrix_simulator.py
```diff
@@ -12,24 +12,38 @@
...
-from braket.default_simulator.jaqcd_simulator import BaseLocalJaqcdSimulator
+from braket.default_simulator.jaqcd_density_matrix_simulator import JaqcdDensityMatrixSimulator
+from braket.default_simulator.openqasm._helpers.utils import singledispatchmethod
+from braket.default_simulator.openqasm_density_matrix_simulator import (
+    OpenQASMDensityMatrixSimulator,
+)
 
 
-class DensityMatrixSimulator(BaseLocalJaqcdSimulator):
+class DensityMatrixSimulator(OpenQASMDensityMatrixSimulator, JaqcdDensityMatrixSimulator):
```
speller26, line 33: A bit concerned about the multiple inheritance here (an in general, to be fair; probably makes my thoughts on OOP obvious). In particular, we're running into the [diamond problem](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem) here, since both parent classes subclass [BaseLocalSimulator](https://github.com/aws/amazon-braket-default-simulator-python/blob/local-sim-branch/src/braket/default_simulator/simulator.py#L55).

Can we do some kind of dispatch instead?

src/braket/default_simulator/density_matrix_simulator.py
```diff
@@ -120,9 +213,21 @@ def properties(self) -> GateModelSimulatorDeviceCapabilities:
...
-                    }
+                    },
                 },
                 "paradigm": {"qubitCount": qubit_count},
                 "deviceParameters": GateModelSimulatorDeviceParameters.schema(),
             }
         )
+
+    @singledispatchmethod
+    def _run_internal(
```
speller26, line 224: Maybe it's just my opinion, but I prefer the default implementation of a `singledispatch` to be either an error or "true" default behavior that doesn't care about the type. In particular, the default implementation should not have a type at all.

src/braket/default_simulator/density_matrix_simulator.py
```diff
@@ -120,9 +213,21 @@ def properties(self) -> GateModelSimulatorDeviceCapabilities:
...
+        self, ir: Union[OQ3Program, JaqcdProgram], *args, **kwargs
```
speller26, line 225: The OQ branch should only accept `OQ3Program`, not `JaqcdProgram`

src/braket/default_simulator/density_matrix_simulator.py
```diff
@@ -120,9 +213,21 @@ def properties(self) -> GateModelSimulatorDeviceCapabilities:
...
+    ) -> GateModelTaskResult:
+        """run OpenQASM program"""
+        return OpenQASMDensityMatrixSimulator.run(self, ir, *args, **kwargs)
```
speller26, line 228: I'm a bit confused; why is it necessary to subclass both `OpenQASMDensityMatrixSimulator` and `JaqcdDensityMatrixSimulator` if `run` just does this dispatch?

src/braket/default_simulator/state_vector_simulator.py
```diff
@@ -127,12 +139,92 @@ def properties(self) -> GateModelSimulatorDeviceCapabilities:
...
                         ],
-                    }
+                    },
+                    "braket.ir.openqasm.program": {
+                        "actionType": "braket.ir.openqasm.program",
+                        "version": ["1"],
+                        "supportedOperations": [
+                            # OpenQASM primitives
+                            "U",
+                            "GPhase",
```
speller26, line 149: Is this also true for JAQCD? I'm not sure `properties` actually needs to be implemented at this base level.

speller26 (commented):
Initial comments, likely more to come

src/braket/default_simulator/density_matrix_simulator.py
ajberdy: Ya I vibe with that, I'll change this

src/braket/default_simulator/density_matrix_simulator.py
ajberdy: Will fix with the default change

src/braket/default_simulator/density_matrix_simulator.py
ajberdy: because we are passing in `self` to the `OpenQASMDensityMatrixSimulator` method. We can't do this if `self` isn't an instance of `OpenQASMDensityMatrixSimulator`

src/braket/default_simulator/state_vector_simulator.py
ajberdy: No, that's why it's only listed under `braket.ir.openqasm.program` and not `braket.ir.jaqcd.program`. Because of the class structure, `StateVectorSimulator` is a subclass to the IR-specific simulators and not the other way around. So this isn't a "base level", it's actually the sub-est class. (more on this in response to your diamond structure comment)

src/braket/default_simulator/density_matrix_simulator.py
ajberdy: Yes, that's how I initially implemented this, but I thought multiple inheritance was a bit neater. Maybe we can chat offline to see what makes more sense

src/braket/default_simulator/simulator.py
```diff
@@ -53,16 +59,79 @@
...
 
 class BaseLocalSimulator(BraketSimulator):
-    @property
-    @abstractmethod
-    def device_action_type(self) -> DeviceActionType:
-        """DeviceActionType"""
-
-    @abstractmethod
+    @singledispatchmethod
     def run(
```
speller26, line 63: Rather than making `run` a `singledispatchmethod`, create a private `_run` method for dispatching instead; public functions marked with `singledispatch` end up producing really ugly docs (you end up with one extra run method, but with blank documentation, for each implementation)

ajberdy: Because of the differing signatures for the run methods of each IR, I kept the signature of the top-level run method general, with a detailed doc string going more in depth about the specific parameters used. I also made the ir-specific run methods public so that the detailed documentation for each ir type would be generated.

speller26 (approved):

PR #101, opened 2022-08-03
fix: basis rotations bug fix

ajberdy:
*Issue #, if available:*

*Description of changes:*

Fix basis rotation bug

*Testing done:*

unit

## Merge Checklist

#### General

#### Tests


src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -63,28 +64,31 @@ def num_qubits(self) -> int:
...
-                    diagonalizing_gates = braket_obs.diagonalizing_gates(self.num_qubits)
-                    basis_rotation_instructions.extend(diagonalizing_gates)
-                    measured_qubits |= set(actual_targets)
+                braket_obs = _from_braket_observable(observables, actual_targets)
+
+                if isinstance(braket_obs, TensorProduct):
+                    for factor in braket_obs.factors:
+                        current_value = observable_map[factor.measured_qubits[0]]
+                        if current_value is not None and type(current_value) != type(factor):
+                            raise ValueError("Qubits not simultaneously measurable")
```
speller26, line 80: 1. Need to account for identity observables; they can _always_ be measured simultaneously
2. What about arbitrary Hermitians? IIRC we do allow simultaneous measurement of the same Hermitian matrix.

src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -63,28 +64,31 @@ def num_qubits(self) -> int:
...
+                        observable_map[factor.measured_qubits[0]] = factor
                 else:
-                    # this shouldn't impact any circuits coming from the BDK
-                    raise NotImplementedError("Partially measured observable target")
+                    current_value = observable_map[braket_obs.measured_qubits[0]]
+                    if current_value is not None and type(current_value) != type(braket_obs):
+                        raise ValueError("Qubits not simultaneously measurable")
+                    observable_map[braket_obs.measured_qubits[0]] = braket_obs
```
speller26, line 86: Considering the need to consider identity observables and Hermitians, this logic (and the logic above that these lines repeat) should be extracted to a private function

src/braket/default_simulator/openqasm/circuit.py
ajberdy: Yup, just posted the first iteration and then went on to add hermitian support

src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -63,28 +64,47 @@ def num_qubits(self) -> int:
...
     def basis_rotation_instructions(self):
-        """
-        This function assumes all observables are commuting.
-        """
-
         basis_rotation_instructions = []
-        measured_qubits = set()
+        observable_map = {}
+
+        def process_observable(observable):
```
speller26, line 70: Why a nested function instead of a private function?

src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -63,28 +64,47 @@ def num_qubits(self) -> int:
...
+                            target == measured_qubits
+                            and type(previously_measured) == type(observable)
+                        ):
+                            conflicts = True
+                        elif (
+                            isinstance(observable, Hermitian)
+                            and isinstance(previously_measured, Hermitian)
+                            and not np.allclose(observable.matrix, previously_measured.matrix)
+                        ):
+                            conflicts = True
```
speller26, line 88: Why not just throw errors directly, and include more information about _why_ they occurred?

src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -63,28 +64,47 @@ def num_qubits(self) -> int:
...
+                        conflicts = False
+                        if not (
+                            target == measured_qubits
+                            and type(previously_measured) == type(observable)
+                        ):
+                            conflicts = True
+                        elif (
+                            isinstance(observable, Hermitian)
+                            and isinstance(previously_measured, Hermitian)
+                            and not np.allclose(observable.matrix, previously_measured.matrix)
```
speller26, line 86: Also need to ensure that the target qubits are in the same order for a multi-qubit observable. For example, a 2-qubit observable is generally not simultaneously measurable on targets [0. 1] and [1, 0].

src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -63,28 +64,47 @@ def num_qubits(self) -> int:
...
+            if isinstance(observable, Identity):
+                return
+            measured_qubits = tuple(observable.measured_qubits)
+            for qubit in measured_qubits:
+                for target, previously_measured in observable_map.items():
+                    if qubit in target:
+                        conflicts = False
+                        if not (
+                            target == measured_qubits
+                            and type(previously_measured) == type(observable)
```
speller26, line 80: This only works now because the only predefined observables are 1q Paulis; it might be worth future-proofing for if we get predefined multi-qubit observables.

src/braket/default_simulator/openqasm/circuit.py
ajberdy: This way it has access to observable map as a closure. I can move it out and add an arg if you're opinionated here though

src/braket/default_simulator/openqasm/circuit.py
ajberdy: Sure, I can do that. I was just keeping the level of granularity like the BDK

src/braket/default_simulator/openqasm/circuit.py
ajberdy: Moved around the logic a bit to make it cleaner. I believe it should be easier to follow and should be general for larger target standard observables now

src/braket/default_simulator/openqasm/circuit.py
ajberdy: My code should be doing that now. The targets are being checked as tuples, so order counts

speller26: D'oh, somehow simultaneously missed that for this comment and yet caught it for the next one

speller26 (approved):

PR #166, opened 2023-06-12
Cirq circuit support

Sai-prakash15:
*Issue #, if available:*

*Description of changes:*

*Testing done:*
Tested all the current test cases. And added new unit tests.

- [x]  Write unit tests

## Merge Checklist

#### General

#### Tests


src/braket/default_simulator/cirq/program_context.py
```diff
@@ -0,0 +1,88 @@
...
+        qubits = [cirq.LineQubit(int(qubit)) for qubit in target]
+        target_qubits = qubits[len(ctrl_modifiers) :]
+        control_qubits = qubits[: len(ctrl_modifiers)]
+        try:
+            if params:
+                gate = CIRQ_GATES[gate_name](*params).on(*target_qubits)
+            else:
+                gate = CIRQ_GATES[gate_name](*target_qubits)
+        except KeyError:
+            raise NotImplementedError
```
ajberdy, line 63: Can we replace this try/except with an if statement checking if the gate is supported?

src/braket/default_simulator/cirq/program_context.py
```diff
@@ -0,0 +1,88 @@
...
+
+        gate = gate.controlled_by(*control_qubits, control_values=ctrl_modifiers)
+        gate = gate**power
```
ajberdy, line 66: How does this behave when power is float? It would be a good idea to include that case when adding unit tests

src/braket/default_simulator/cirq/program_context.py
```diff
@@ -0,0 +1,88 @@
...
+    # "zz": cirq.ZZ,
+    "ccnot": cirq.CCNOT,
+    "cswap": cirq.CSWAP,
+    "ccz": cirq.CCZ,
+    "ccx": cirq.CCX,
+    "measure": cirq.MeasurementGate,
+}
+
+
+class ProgramContext(AbstractProgramContext):
```
ajberdy, line 37: Big fan of the structure here, very cool that you can do this without touching the interpreter

src/braket/default_simulator/cirq/program_context.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def add_gate_instruction(
+        self, gate_name: str, target: Tuple[int], params, ctrl_modifiers: List[int], power: int
+    ):
+        qubits = [cirq.LineQubit(int(qubit)) for qubit in target]
+        target_qubits = qubits[len(ctrl_modifiers) :]
+        control_qubits = qubits[: len(ctrl_modifiers)]
+        try:
+            if params:
+                gate = CIRQ_GATES[gate_name](*params).on(*target_qubits)
```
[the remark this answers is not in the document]
Sai-prakash15: When the gates accept an angle parameter. Few gates were already initialized. I was using them. I could use on in both the cases.

Let me change it quickly to avoid confusion

src/braket/default_simulator/cirq/program_context.py
```diff
@@ -0,0 +1,88 @@
...
+    "t": cirq.T,
+    "cphaseshift": cirq.cphase,
+    "rx": cirq.rx,
+    "ry": cirq.ry,
+    "rz": cirq.rz,
+    "swap": cirq.SWAP,
+    "iswap": cirq.ISWAP,
+    # "xx": cirq.XX,
+    # "yy": cirq.YY,
+    # "zz": cirq.ZZ,
```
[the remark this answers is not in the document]
ajberdy: I think that should still be fine. Two gates (braket / cirq) with the same name should work on their respective simulators at the same time. Just here, no parameters will be expected.

However, it may be confusing to customers trying to use the xx gate from the BDK, since that refers to the Braket xx gate specifically. I'm ok with leaving these out for now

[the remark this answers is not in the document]
Sai-prakash15: https://github.com/aws/amazon-braket-default-simulator-python/pull/166/files#r1227109496 

I did not see the comment when I put he above comment. I don't think they use web sockets. Haha

src/braket/default_simulator/cirq/program_context.py
Sai-prakash15: Makes sense. 

There was already a check if a gate is defined or not while calling this function. So I removed the check completely

src/braket/default_simulator/cirq/program_context.py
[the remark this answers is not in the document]
Sai-prakash15: Since we are focusing to convert a braket specific openqasm strings. 
Even if I add a conditional check for these gates to make an exception to construct the circuit it would not be equivalent to the input braket circuit.

Do I just comment them or remove them all together similar to other gates which I was not able to find in the documentation.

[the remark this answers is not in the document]
ajberdy: I would remove them and add a comment above `CIRQ_GATES` noting that xx, yy, and zz are not the same gates between Braket and Cirq

src/braket/default_simulator/cirq/program_context.py
```diff
@@ -0,0 +1,84 @@
...
+        self, gate_name: str, target: Tuple[int], params, ctrl_modifiers: List[int], power: int
+    ):
+        qubits = [cirq.LineQubit(int(qubit)) for qubit in target]
+        target_qubits = qubits[len(ctrl_modifiers) :]
+        control_qubits = qubits[: len(ctrl_modifiers)]
+
+        if params:
+            gate = CIRQ_GATES[gate_name](*params).on(*target_qubits)
+        else:
+            gate = CIRQ_GATES[gate_name].on(*target_qubits)
```
ajberdy, line 59: would this line work the same as `gate = CIRQ_GATES[gate_name]().on(*target_qubits)`?

If so, maybe we can remove the if/else and add a line `params = params or []` and then just do `gate = CIRQ_GATES[gate_name](*params).on(*target_qubits)`

Sai-prakash15: I was thinking to do it exactly like you said.

The problem is gates with params have __init__ which accepts angle. And other gates which use different gates to define the gate are already initialized. 

Objects have __call__ method which accepts qubits.

Hence I decided to go with if condition. Let me know if you think we can find a way around this to just write a single line.

test/unit_tests/braket/default_simulator/cirq/test_openqasm_conversion.py
```diff
@@ -0,0 +1,216 @@
...
+
+from braket.default_simulator.cirq.program_context import ProgramContext
+from braket.default_simulator.openqasm.interpreter import Interpreter
+
+
+@pytest.mark.parametrize(
+    "openqasm",
+    (
+        (
+            """
```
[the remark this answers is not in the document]
ajberdy: This is the correct approach. The default simulator should not use the BDK since it is a dependency of the BDK and that would create a circular import.

src/braket/default_simulator/cirq/program_context.py
ajberdy: this should be fine then

test/unit_tests/braket/default_simulator/cirq/test_openqasm_conversion.py
```diff
@@ -0,0 +1,216 @@
...
+        """
+        ),
+    ),
+)
+def test_openqasm_conversion_to_cirq(openqasm):
+    circuit = Interpreter(ProgramContext()).build_circuit(
+        source=openqasm,
+    )
+
+    assert isinstance(circuit, cirq.Circuit)
```
ajberdy, line 216: can we add a correctness check here? Maybe we can compare the state vectors of the two circuits?

Sai-prakash15: That would mean I have to run it on a simulator right?

I was thinking to do it and then if I would directly access a cirq simulator here. But i was thinking to create a plugin like below in the end once I complete the implementation of all the available gates.
I have to create a PR to add support for localSimulator("cirq") in the sdk repo. 

Do you think it is better to do it now by directly accessing the cirq simulator  and refactor it later to use the simulator via the plugin.

ajberdy: `LocalSimulator("cirq")` is an entirely separate consideration, since it lives in the BDK and we won't be referencing BDK code at all in this repo. Take a look at [this test](https://github.com/aws/amazon-braket-default-simulator-python/blob/main/test/unit_tests/braket/default_simulator/openqasm/test_interpreter.py#L1006) for an example of how to simulate a circuit a local sim unit test. For cirq, you can extract the state vector using its library methods, maybe something like [this](https://quantumai.google/reference/python/cirq/final_state_vector)

Sai-prakash15: Thank you.

I just have a small query about how would I know what the expected vector is?

I have generated the openQasm strings from the Braket circuits.

Do I have do something like this
`simulation = StateVectorSimulation(2, 1, 1)
    simulation.evolve(circuit.instructions)`
  to obtain a expected value ?
   Or is there a better way and more accurate way of doing it
   
   
For example:
` 
circ = BCircuit.Circuit().cnot(control=0, target=1, power =0.5)
circuit = Interpreter(ProgramContext()).build_circuit(
    source=circ.to_ir(IRType.OPENQASM).source,
)
 cirq.final_state_vector(circuit)
 `
  
  Output: 
`[1.+0.j 0.+0.j 0.+0.j 0.+0.j]`  
 
In a nut shell my query is what is the best way to know the expected value and compare it to the above value since I dont know how to make the calculations manually.

Sai-prakash15: I was thinking of it but I was not sure. Thanks for the references!

Sai-prakash15: This caught a bug.

ctrl_modifiers = [bit^1 for bit in ctrl_modifiers]

Along with the added check,  fixed the bug. Nice catch : )

test/unit_tests/braket/default_simulator/cirq/test_openqasm_conversion.py
```diff
@@ -0,0 +1,279 @@
...
+                0.0 + 0.0j,
+                0.00417738 + 0.00324475j,
+                0.0 + 0.0j,
+                -0.00417738 - 0.00324475j,
+                0.0 + 0.0j,
+            ],
+        ),
+    ),
+)
+def test_openqasm_conversion_to_cirq(openqasm, expected_result):
```
ajberdy, line 273: How come we're passing in expected result as an argument? It seems it should be pretty easy to generate from the OpenQASM in the test

Sai-prakash15: I did not know we can generate directly final state vector from OpenQASM.
Could you send me a reference?

I generated all the expected result using the state vector simulator

ajberdy: Yes exactly, I'm saying you can do the same thing (as I linked in another comment) in the test

Sai-prakash15: I thought that would create a dependency. But it makes sense for future testcases

test/unit_tests/braket/default_simulator/cirq/test_openqasm_conversion.py
```diff
@@ -0,0 +1,218 @@
...
+    ),
+)
+def test_openqasm_conversion_to_cirq(openqasm):
+    circuit = Interpreter(ProgramContext()).build_circuit(
+        source=openqasm,
+    )
+
+    braket_circuit = Interpreter().build_circuit(
+        source=openqasm,
+    )
```
ajberdy, line 213: These lines appear to be identical, except for the ProgramContext argument, which looks like the default value. I understand they're not because it's actually the cirq/ProgramContext object, but maybe we should rename to `CirqProgramContext` to differentiate? Also nit-- cirq_circuit and braket_circuit might be slightly clearer

test/unit_tests/braket/default_simulator/cirq/test_openqasm_conversion.py
```diff
@@ -0,0 +1,218 @@
...
+            cphaseshift(0.15) q[2], q[4];
+            swap q[3], q[4];
+            ccnot q[2], q[3], q[4];
+            cswap q[1], q[3], q[4];
+            b[0] = measure q[0];
+            b[1] = measure q[1];
+            b[2] = measure q[2];
+            b[3] = measure q[3];
+            b[4] = measure q[4];
+            """
```
ajberdy, line 202: small nit: maybe we could have consistent indentation of all the strings here?

Sai-prakash15: I thought black would do it but no : (

ajberdy: Yup usually haha

src/braket/default_simulator/cirq/program_context.py
```diff
@@ -0,0 +1,84 @@
...
+    "iswap": cirq.ISWAP,
+    "ccnot": cirq.CCNOT,
+    "cswap": cirq.CSWAP,
+    "ccz": cirq.CCZ,
+    "ccx": cirq.CCX,
+    "measure": cirq.MeasurementGate,
+}
+
+
+class CirqProgramContext(AbstractProgramContext):
```
ajberdy, line 35: last comment I think-- the file name should be updated to correspond to the class name

Sai-prakash15: It's alright. You are welcome to post as many comments in the last minute. 
I agree too.

ajberdy (approved):

PR #174, opened 2023-06-28
feat: `AbstractProgramContext` interface

Sai-prakash15:
*Issue #, if available:*

*Description of changes:*
Abstract the program context to define the functions which are necessary to build a circuit.
Add properties: probability, probabilities and gamma to noise operators.

*Testing done:*
Add unit test and Tested against all the existing test cases
## Merge Checklist

#### General

#### Tests


src/braket/default_simulator/openqasm/parser/braket_pragmas.py
```diff
@@ -213,7 +213,7 @@ def visitProbabilities(self, ctx: BraketPragmasParser.ProbabilitiesContext):
         return [float(prob.symbol.text) for prob in ctx.children[::2]]
 
 
-def parse_braket_pragma(pragma_body: str, qubit_table: "QubitTable"):
+def parse_braket_pragma(pragma_body: str, pragma_node_visitor=None):
```
ajberdy, line 216: What's the motivation for this change? I don't think program context should need to import and instantiate a pragma node visitor, rather than just passing the qubit table

Sai-prakash15: When I made the pragma node visitor abstract I changed it to accept pragma node visitor.
I felt like maybe in the future if someone wants to modify the behavior like this.

I will change it back to accept qubit table. I don't think anyone would change the behavior of pragma node visitor from program context

src/braket/default_simulator/openqasm/interpreter.py
```diff
@@ -608,7 +608,7 @@ def handle_builtin_gate(
         qubits: List[Union[Identifier, IndexedIdentifier]],
         modifiers: List[QuantumGateModifier],
     ) -> None:
-        """Add unitary operation to the circuit"""
+        """Add unitary operation to the program"""
```
ajberdy, line 611: I don't think I agree with this change. Here "Program" refers to the object being interpreted and "Circuit" the object being constructed; the operation is being added to the latter

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -375,7 +377,7 @@ class ProgramContext:
 
     """
 
-    def __init__(self):
+    def __init__(self, program=Circuit()):
```
ajberdy, line 380: Similar to above, why not name this `circuit`?

src/braket/default_simulator/openqasm/parser/generated/BraketPragmasParser.py
ajberdy, line 1: We probably shouldn't be updating generated files

src/braket/default_simulator/simulator.py
```diff
@@ -345,7 +346,7 @@ def run_openqasm(
                 are requested when shots>0.
         """
         is_file = openqasm_ir.source.endswith(".qasm")
-        interpreter = Interpreter()
+        interpreter = Interpreter(context=ProgramContext())
```
ajberdy, line 349: This shouldn't be necessary, right? I'd like `Interpreter()` to work as expected with default arg values

src/braket/default_simulator/openqasm/interpreter.py
Sai-prakash15: I was thinking the same but Kshitij told me it could also support pulse programs in the future. And I did not know much about pulse programs. Hence changed the name to program

https://github.com/aws/amazon-braket-default-simulator-python/pull/162

Should I just change it back to circuit or leave it ?

ajberdy: > OpenQASM encapsulates more information than what is described by the circuit class today. So it might be helpful to call this argument Program and pass in a circuit (which is a subset of a program)

The "circuit" here refers to what is built by the interpreter, which with today's architecture, is a level of functionality that we've been calling a circuit. Updates to the simulator to support more of a "program-level" scope are likely to involve a different architecture (ie. not building a single interpreted object). For example the [prototype program simulator](https://github.com/aws/amazon-braket-default-simulator-python/commit/38853b3edac022c9e16c7b7b04b7880ad9ec377a#diff-2e92ff2ca1084c457cd127f2a76a969d0e404853f4f5d74db8a0c8245267ce16R400) doesn't utilize the "circuit-building" part of program context, except for temporarily storing a circuit-level instruction buffer.

Given that context, I think `circuit` makes more sense. Feel free to weight in here @kshitijc

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -571,6 +579,44 @@ def add_builtin_gate(
...
                 power *= mod.argument.value
+        self.add_gate_instruction(
+            gate_name, target, params, ctrl_modifiers=ctrl_modifiers, power=power
+        )
+
+    @abstractmethod
+    def add_gate_instruction(
+        self, gate_name: str, target: Tuple[int], params, ctrl_modifiers: List[int], power: int
+    ):
+        """Add gate instruction to the circuit"""
```
speller26, line 590: Also add argument information:
```
"""
Add gate instruction to the circuit.

Args:
    gate_name (str): ...
    target (int): ...
"""
```
Do this for other methods too

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -375,7 +376,7 @@ class ProgramContext:
 
     """
 
-    def __init__(self):
+    def __init__(self, circuit=Circuit()):
```
speller26, line 379: Rather than having this in the base constructor, leave it to the subclasses.
```suggestion
    def __init__(self):
```

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -384,7 +385,7 @@ def __init__(self):
         self.scope_manager = ScopeManager(self)
         self.inputs = {}
         self.num_qubits = 0
-        self.circuit = Circuit()
+        self.circuit = circuit
```
speller26, line 388: Remove this altogether...
```suggestion
```

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -571,6 +579,44 @@ def add_builtin_gate(
...
+        """Add a custom Unitary instruction to the circuit"""
+
+    @abstractmethod
+    def add_noise_instruction(self, *args, **kwargs):
+        """Add a noise instruction to the circuit"""
+
+
+class ProgramContext(AbstractProgramContext):
+    def __init__(self):
+        super().__init__(Circuit())
```
speller26, line 607: ... And set the circuit here
```suggestion
    def __init__(self):
        super().__init__()
        self._circuit = Circuit()
```

src/braket/default_simulator/openqasm/program_context.py
ajberdy: I don't think I agree with this. We could make the default None, but caching the "starting circuit" to `self.circuit` could totally be base class behavior. This is more than just a type annotation, it allows users to start not from scratch

src/braket/default_simulator/openqasm/program_context.py
ajberdy: Maybe this could be
```suggestion
    def __init__(self, circuit=None):
```

src/braket/default_simulator/openqasm/program_context.py
ajberdy: This behavior is likely duplicated in all subclasses, so to me it would make sense to keep that assignment in the base class

src/braket/default_simulator/openqasm/program_context.py
speller26: But rather than keeping it in the abstract class, how about letting subclasses decide that? It feels kinda weird having the parent class decide that all base classes will build up a circuit _object_.

ajberdy: Hm I suppose that's true, just all the ones so far haha

src/braket/default_simulator/openqasm/program_context.py
speller26: Is that necessarily the case though? I can totally conceive of a situation where an implementation looks like
```python
class MyContext(AbstractProgramContext):
    def __init__(self, ...):
        self._a = ...
        self._b = ...

    def exec(self):
        self._a.do_something()
        self._b.do_something()
```

src/braket/default_simulator/openqasm/program_context.py
ajberdy: Even the monte-carlo/mid-circuit measurement sim has a circuit object in the program context, but that's just a circuit-level buffer of instructions that gets popped every time a measurement happens

src/braket/default_simulator/openqasm/program_context.py
ajberdy: The interpreter class does have a `build_circuit` function, which returns `context.circuit`, but other than that there is no direct coupling between the interpreter and circuit classes. If someone wanted to build an interpreter/simulator that didn't have a build_circuit method, then yup I agree

speller26: Ah, in that case, it does make sense then.

src/braket/default_simulator/openqasm/program_context.py
speller26: Yep, don't want to have mutable defaults.

src/braket/default_simulator/openqasm/program_context.py
ajberdy: Arguably, if we carry on this line of thought, an abstract interpreter class wouldn't have a build_circuit class. For example the mcm interpreter doesn't utilize that method at all, but has a [simulate](https://github.com/aws/amazon-braket-default-simulator-python/blob/mcm-sim/src/braket/default_simulator/openqasm/native_interpreter.py#L39) method instead

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -375,7 +387,7 @@ class ProgramContext:
 
     """
 
-    def __init__(self):
+    def __init__(self, program=Circuit()):
```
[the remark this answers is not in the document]
Sai-prakash15: I did not know that this was the reason.

Thanks for sharing it!

src/braket/default_simulator/openqasm/program_context.py
speller26: Looking at `Interpreter`, it seems pretty deeply linked to the default simulator's `Circuit` class; no need to do it in this PR, but it seems to make sense to the visitor functionality from the build/run functionality.

speller26 (approved):

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -571,6 +745,68 @@ def add_builtin_gate(
...
+                                        of the noise operation. (Optional)
+                - `probabilities` (List[float]): The probabilities associated with each possible outcome
+                                                of the noise operation. (Optional)
+        """
+
+
+class ProgramContext(AbstractProgramContext):
+    def __init__(self):
+        super().__init__()
+        self.circuit = Circuit()
```
ajberdy, line 797: if this specific program context class has a `self.circuit` attribute, its init should also accept a circuit argument of the same type

speller26: Quick fix: #183

Issue #176, opened 2023-06-29
Out of bounds error in SciPy solver


johnzl-777:
**Describe the bug**
Trying to run the following task:
```python
from braket.devices import LocalSimulator
from braket.ir.ahs import Program

task_json = {
  "braketSchemaHeader": {
    "name": "braket.ir.ahs.program",
    "version": "1"
  },
  "setup": {
    "ahs_register": {
      "sites": [
        [
          "0E-8",
          "0E-8"
        ],
        [
          "0.00000610",
          "0E-8"
        ],
        [
          "0.00001220",
          "0E-8"
        ],
        [
          "0.00001830",
          "0E-8"
        ],
        [
          "0.00002440",
          "0E-8"
        ],
        [
          "0.00003050",
          "0E-8"
        ],
        [
          "0.00003660",
          "0E-8"
        ],
        [
          "0.00004270",
          "0E-8"
        ],
        [
          "0.00004880",
          "0E-8"
        ],
        [
          "0.00005490",
          "0E-8"
        ],
        [
          "0.00006100",
          "0E-8"
        ]
      ],
      "filling": [
        1,
        1,
        1,
        1,
        1,
        1,
        1,
        1,
        1,
        1,
        1
      ]
    }
  },
  "hamiltonian": {
    "drivingFields": [
      {
        "amplitude": {
          "time_series": {
            "values": [
              "0.0",
              "15700000.0",
              "15700000.0",
              "0.0"
            ],
            "times": [
              "0E-9",
              "3.00E-7",
              "6.00E-7",
              "6.60E-7"
            ]
          },
          "pattern": "uniform"
        },
        "phase": {
          "time_series": {
            "values": [
              "0E-7",
              "0E-7"
            ],
            "times": [
              "0E-9",
              "6.60E-7"
            ]
          },
          "pattern": "uniform"
        },
        "detuning": {
          "time_series": {
            "values": [
              "-18800000.0",
              "-18800000.0",
              "-12218750.0",
              "0.0"
            ],
            "times": [
              "0E-9",
              "3.00E-7",
              "6.00E-7",
              "6.60E-7"
            ]
          },
          "pattern": "uniform"
        }
      }
    ],
    "shiftingFields": []
  }
}

LocalSimulator("braket_ahs").run(Program(**task_json), shots = 10)
```
produces the following error:
```
capi_return is NULL
Call-back cb_f_in_zvode__user__routines failed.
[...]

Issue #168, opened 2023-05-30
Numerical issue with the local AHS simulator


balewski:
**Describe the bug**
Braket emulator crashes  for a system of 10 atoms, with the error:
  File "/usr/local/lib/python3.10/dist-packages/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py", line 566, in _apply_hamiltonian
    output_register += (rabi_coef[index_time] / 2) * rabi_op.dot(input_register)
IndexError: index 400 is out of bounds for axis 0 with size 400

**To Reproduce**
execute attached demonstrator as-is. 
https://bitbucket.org/balewski/quantummind/src/master/QuEra/python/issues/issue1_10atom_crash.py
It will crash in 5 seconds.

But if you reduce the number of atoms in line 74 to 9
 nAtom=9
all works fine.

**Expected behavior**
program should simulate up to nAtom=16

**System information (please complete the following information as applicable):**
$ pip3 list |grep braket
amazon-braket-default-simulator 1.14.0
amazon-braket-schemas           1.17.0
amazon-braket-sdk               1.39.1

  - Working Environment: podman image running locally

shpface:
Thanks for the bug report. I have transferred the issue to the default simulator repository as that is where the bug is. There seems to be an issue with the usage of the ode integrator from scipy. We will look into the source of the bug.

shpface:
Hi Jan, this bug is due to the integrator taking intermediate steps outside of the range of interpolated values that are computed before the integration starts. I'll take a look at what needs to be changed to correct this error. In the meantime, can you try using `device.run(..., solver_method="bdf")` to select a different integration method and see if that one avoids this bug?

balewski:
Hi Stephen,
yes, I confirm changing solver avoids this crash.
Now I can run my simulations on a bigger system. It is a big help!
Thank you for investigating it
Jan

guomanmin:
This issue seems resolved, closing

PR #180, opened 2023-06-30
Mao/fix_scipy_bug

maolinml:
*Issue #, if available:*

There is a bug flagged by two customers in https://github.com/aws/amazon-braket-default-simulator-python/issues/168 and https://github.com/aws/amazon-braket-default-simulator-python/issues/176. In particular, the intermediate time index used by the scipy integrator may exceed the limit.

*Description of changes:*
When the intermediate time index exceeds the limit, we use the very last time index for approximation, and issue a warning for that.

*Testing done:*

## Merge Checklist

#### General

- [ ] I have read the [CONTRIBUTING](https://github.com/aws/amazon-braket-default-simulator-python/blob/main/CONTRIBUTING.md) doc
- [ ] I used the commit message format described in [CONTRIBUTING](https://github.com/aws/amazon-braket-default-simulator-python/blob/main/CONTRIBUTING.md#commit-your-change)
- [ ] I have updated any necessary documentation, including [READMEs](https://github.com/aws/amazon-braket-default-simulator-python/blob/main/README.md) and [API docs](https://github.com/aws/amazon-braket-default-simulator-python/blob/main/CONTRIBUTING.md#documentation-guidelines) (if appropriate)

#### Tests

### DIFF
--- a/src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
+++ b/src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
@@ -495,6 +495,12 @@ def _get_hamiltonian(
     ) = operators_coefficients
 
     index_time = int(index_time)
+
+    # If the integrator uses intermediate time value outside of the specified time range,
+    # the final time value is used as an approximation.
+    if index_time >= len(rabi_coefs):
+        index_time = len(rabi_coefs) - 1
+
     hamiltonian = interaction_op
 
     # Add the driving fields
@@ -557,6 +563,12 @@ def _apply_hamiltonian(
     ) = operators_coefficients
 
     index_time = int(index_time)
+
+    # If the integrator uses intermediate time value outside of the specified time range,
+    # the final time value is used as an approximation.
+    if index_time >= len(rabi_coefs):
+        index_time = len(rabi_coefs) - 1
+
     output_register = interaction_op.dot(input_register)
 
     # Add the driving fields
--- a/src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
+++ b/src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
@@ -455,7 +455,7 @@ def _print_progress_bar(num_time_points: int, index_time: int, start_time: float
 
 
 def _get_hamiltonian(
-    index_time: int,
+    index_time: float,
     operators_coefficients: Tuple[
         List[scipy.sparse.csr_matrix],
         List[scipy.sparse.csr_matrix],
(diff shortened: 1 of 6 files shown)


src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
```diff
@@ -495,6 +496,18 @@ def _get_hamiltonian(
     ) = operators_coefficients
 
     index_time = int(index_time)
+
+    # If the integrator uses intermediate time value outside of the specified time range,
+    # the final time value is used as an approximation.
+    if index_time >= len(rabi_coefs):
+        index_time = len(rabi_coefs) - 1
```
shpface, line 503: This will still be out of bounds if `index_time` is more than 1 greater than the length. Also we should handle the case of the integrator sampling before the start of the first interval. Let's also change the type hint on `index_time` to `float`.

src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
```diff
@@ -557,6 +570,18 @@ def _apply_hamiltonian(
...
```
shpface, line 577: Same comments as in `_get_hamiltonian`.

src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
maolinml: 1. If `index_time` is more than 1 larger, it will be fixed at `len(rabi_coefs) - 1` so it should be good right?
2. I have add another checking for the `index_time <0`. 
3. I change the hint for `index_time` to `float`.

shpface: 1. You are right. I misread what you had there. All good.
2. Rather than branching, if you use `max(0, min(len - 1, index_time))`, you can avoid the additional lines that are missing test coverage

src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
```diff
@@ -495,6 +496,29 @@ def _get_hamiltonian(
     ) = operators_coefficients
 
     index_time = int(index_time)
+
+    # If the integrator uses intermediate time value that is larger than the maximum
+    # time value specified, the final time value is used as an approximation.
+    if index_time >= len(rabi_coefs):
+        index_time = len(rabi_coefs) - 1
```
yzbAmazon, line 503: I think another option is to return zero coefficient for any index time larger than maximum. Then, the Hamiltonian becomes zero after the maximum time and you can safely return something for extended index time without introducing possible errors.

maolinml:
Many thanks to both of you for taking a look at this PR. I added a new unit test for `_get_hamiltonian` and `_apply_hamiltonian` to recover the coverage. @yzbAmazon Yes, that is another option, but I am worried that the abrupt change of coefficients may bring in other issues, say the integrator can't converge, or the result is incorrect. I hope that using the last index as an approximation, or performing some kind of extrapolation may avoid these issues. Let me know if you have other thoughts.

shpface (approved):

PR #199, opened 2023-08-03
fix: Support `angle` declarations

speller26:
Also added license headers to some files

*Issue #, if available:*

https://github.com/aws/amazon-braket-sdk-python/issues/632

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/openqasm/_helpers/casting.py
+++ b/src/braket/default_simulator/openqasm/_helpers/casting.py
@@ -1,3 +1,16 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
 import warnings
 from copy import deepcopy
 from functools import singledispatch
@@ -7,6 +20,7 @@ import numpy as np
 from sympy import Symbol
 
 from ..parser.openqasm_ast import (
+    AngleType,
     ArrayLiteral,
     ArrayType,
     BitstringLiteral,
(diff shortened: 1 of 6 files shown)


src/braket/default_simulator/openqasm/_helpers/casting.py
```diff
@@ -99,8 +113,9 @@ def _(into: UintType, variable: LiteralType) -> IntegerLiteral:
     return IntegerLiteral(value)
 
 
-@cast_to.register
-def _(into: FloatType, variable: LiteralType) -> FloatLiteral:
+@cast_to.register(FloatType)
+@cast_to.register(AngleType)
+def _(into, variable: LiteralType) -> FloatLiteral:
```
ajberdy, line 118: The no size case is shared between float and angle, but for specified-bit variables, the behavior is different

ajberdy: I think the way to handle this would be to store fixed-sized angles as `BitstringLiteral`s and update `handle_parameter_value` to do the % 2pi math when building the circuit

ajberdy (dismissed):
The benefit of the angle type is to express fractions of π exactly. We lose this by storing internally as a float literal, making this implementation of angle exactly equivalent to float.

We should likely be keeping these values as fractions of (2) pi and then allowing the backend (ProgramContext, here) to resolve these values into instructions to simulate/execute.

Because an unsized angle has ambiguous behavior, I think storing as a float is fine for today's purposes. Though technically, there is already a regression from a 1-bit angle and 64-bit float in being able to represent π exactly.

src/braket/default_simulator/openqasm/_helpers/casting.py
```diff
@@ -103,14 +117,21 @@ def _(into: UintType, variable: LiteralType) -> IntegerLiteral:
...
+        raise ValueError("Float size must be one of {16, 32, 64}.")
+    value = float(np.array(variable.value, dtype=np.dtype(f"float{into.size.value}")))
     return FloatLiteral(value)
 
 
+@cast_to.register
+def _(into: AngleType, variable: LiteralType) -> FloatLiteral:
+    """Cast angle to float"""
+    if into.size is None:
+        return FloatLiteral(float(variable.value))
```
ajberdy, line 131:
```suggestion
        return FloatLiteral(float(variable.value) % (2 * np.pi))
```

ajberdy (approved):

PR #219, opened 2023-12-15
feat: support symbols as qubit indices

ajberdy:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/openqasm/interpreter.py
+++ b/src/braket/default_simulator/openqasm/interpreter.py
@@ -380,6 +380,8 @@ class Interpreter:
         qubit_lengths = np.array(
             [self.context.qubit_mapping.get_qubit_size(qubit) for qubit in qubits]
         )
+        if any(not isinstance(qubit_length, np.integer) for qubit_length in qubit_lengths):
+            raise NotImplementedError("Undetermined qubit register size")
         register_lengths = qubit_lengths[qubit_lengths > 1]
         if register_lengths.size:
             reg_length = register_lengths[0]
--- a/src/braket/default_simulator/openqasm/program_context.py
+++ b/src/braket/default_simulator/openqasm/program_context.py
@@ -12,13 +12,13 @@
 # language governing permissions and limitations under the License.
 
 from abc import ABC, abstractmethod
-from collections.abc import Iterable
+from collections.abc import Iterable, Sequence
 from functools import singledispatchmethod
 from typing import Any, Optional, Union
 
 import numpy as np
 from braket.ir.jaqcd.program_v1 import Results
-from sympy import Expr
+from sympy import Expr, Symbol
 
 from braket.default_simulator.gate_operations import BRAKET_GATES, GPhase, Unitary
 from braket.default_simulator.noise_operations import (
@@ -47,6 +47,7 @@ from .circuit import Circuit
 from .parser.braket_pragmas import parse_braket_pragma
 from .parser.openqasm_ast import (
     ClassicalType,
+    DiscreteSet,
     FloatLiteral,
     GateModifierName,
     Identifier,
@@ -57,6 +58,7 @@ from .parser.openqasm_ast import (
     QuantumGateModifier,
     RangeDefinition,
     SubroutineDefinition,
+    SymbolLiteral,
 )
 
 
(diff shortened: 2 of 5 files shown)


src/braket/default_simulator/openqasm/interpreter.py
```diff
@@ -380,6 +380,8 @@ def _(self, node: QuantumGate) -> None:
         qubit_lengths = np.array(
             [self.context.qubit_mapping.get_qubit_size(qubit) for qubit in qubits]
         )
+        if any(not isinstance(qubit_length, np.integer) for qubit_length in qubit_lengths):
+            raise NotImplementedError("Undetermined qubit register size")
```
ajberdy, line 384: We run into issues with `cnot qs[0:n1], qs[0:n2];` as we can't verify whether these registers are the same size and furthermore the logic immediately below this won't know how many gates to create and what symbols to use for each.

We can remove this constraint once we support a symbolic "loop" construct in the simulator IR

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -113,29 +115,73 @@ def _(self, identifier: IndexedIdentifier) -> tuple[int]:
         corresponding to the elements referenced by the indexed identifier.
         """
         name = identifier.name.name
+        indices = self.get_qubit_indices(identifier)
+        primary_index = indices[0]
+
+        if isinstance(primary_index, (IntegerLiteral, SymbolLiteral)):
+            target = (self[name][0] + primary_index.value,)
```
ajberdy, line 122: We make small changes to not break when working with symbols.
```
self[name][primary_index.value]
```
turns into 
```
self[name][0] + primary_index.value
```
since we can add a symbol to an int, but can't get a symbol index

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -113,29 +115,73 @@ def _(self, identifier: IndexedIdentifier) -> tuple[int]:
...
+            index_list = convert_discrete_set_to_list(primary_index)
+            target = tuple([self[name][0] + index for index in index_list])
+
+        if len(indices) == 2:
+            # used for gate calls on registers, index will be IntegerLiteral
+            secondary_index = indices[1].value
+            target = (target[secondary_index],)
+
+        # validate indices manually, since we use addition instead of indexing to
+        # accommodate symbolic indices
```
ajberdy, line 136: This function is basically the same as before, with the `[]` -> `+` changes and this extra validation (that we lose from not using indexing). Technically the error we get now is better anyways

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -113,29 +115,73 @@ def _(self, identifier: IndexedIdentifier) -> tuple[int]:
...
+                raise IndexError(
+                    f"Index {relative_index} out of bounds for qubit '{name}' "
+                    f"with size {len(self[name])}"
+                )
+        return target
+
+    @staticmethod
+    def get_qubit_indices(
+        identifier: IndexedIdentifier,
+    ) -> list[IntegerLiteral | RangeDefinition | DiscreteSet]:
```
ajberdy, line 148: This logic used to live inside `get_by_identifier(indexed_identifier)`, but we reuse it for `get_qubit_size` now

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -113,29 +115,73 @@ def _(self, identifier: IndexedIdentifier) -> tuple[int]:
...
-            return (target[secondary_index],)
+            secondary_index = identifier.indices[1][0]
+            return [primary_index, secondary_index]
         else:
             raise IndexError("Cannot index multiple dimensions for qubits.")
 
+    def get_indices_length(
+        self,
+        indices: Sequence[IntegerLiteral | SymbolLiteral | RangeDefinition | DiscreteSet],
+    ):
```
ajberdy, line 167: This method allows us to get the length of a qubit register that may have unbound symbols. This will be more crucial (and better tested) when the range transformation can actually be used with a symbolic "loop" construct in the IR. For now, if this returns something symbolic, the interpreter can't use it.

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -113,29 +115,73 @@ def _(self, identifier: IndexedIdentifier) -> tuple[int]:
...
+            stop = last_index.end.value + buffer
+            step = last_index.step.value if last_index.step is not None else 1
+            return (stop - start) // step
+        elif isinstance(last_index, DiscreteSet):
+            return len(last_index.values)
+
     def get_qubit_size(self, identifier: Union[Identifier, IndexedIdentifier]) -> int:
+        if isinstance(identifier, IndexedIdentifier):
+            indices = self.get_qubit_indices(identifier)
+            return self.get_indices_length(indices)
```
ajberdy, line 184: Get size of qubit without necessarily requiring resolved indices

src/braket/default_simulator/simulator.py
```diff
@@ -351,6 +351,12 @@ def _validate_input_provided(self, circuit: Circuit) -> None:
                     except TypeError:
                         missing_input = param.free_symbols.pop()
                         raise NameError(f"Missing input variable '{missing_input}'.")
+            for qubit in instruction.targets:
+                try:
+                    float(qubit)
+                except TypeError:
+                    missing_input = qubit.free_symbols.pop()
+                    raise NameError(f"Missing input variable '{missing_input}'.")
```
ajberdy, line 359: Also check qubit targets for missing inputs

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -113,29 +115,73 @@ def _(self, identifier: IndexedIdentifier) -> tuple[int]:
...
-        elif isinstance(primary_index, RangeDefinition):
-            target = tuple(np.array(self[name])[convert_range_def_to_slice(primary_index)])
-        # Discrete set
-        else:
-            target = tuple(np.array(self[name])[convert_discrete_set_to_list(primary_index)])
 
         if len(identifier.indices) == 1:
-            return target
+            return [primary_index]
         elif len(identifier.indices) == 2:
```
yitchen-tim, line 157: Do we need to check, like `primary_index`,  if the `secondary_index` is a List and validate its length?

yitchen-tim: Or it's not revenant because it's always an integerliteral?

ajberdy: Technically there's a possibility for something like `qs[0:10][0:5]`, but that is not currently supported. I believe that would throw an error somewhere else, but I could verify. For now, I don't think this is much of a priority. The main (only) place where we see double indices for qubits so far is with `h qs[0:3]` would get mapped to four instructions: `h qs[0:3][0]`, ..., `h qs[0:3][3]`

We could add that validation though anyways

yitchen-tim: I see, I agree that `qs[0:10][0:5]` is an edge use case. We don't need to prioritize the validation I purposed. 

If you were adding the validation, I suggest adding for all elements in `identifier.indices`. I can imagine if `qs[0:10][0:5]` is allowed, users would do `qs[0:10][0:5][1:3]` and so on.

yitchen-tim (approved):

PR #232, opened 2024-03-13
feature: allow support for a subset of measurements

ashlhans:
*Issue #, if available:*

*Description of changes:*

Add support for the OpenQASM `measure` instruction instead of just returning measurements for all qubits

- Add implementation of the `QuantumMeasurement` and `QuantumMeasurmementStatement` in `interpreter.py` to support the OpenQASM measure instruction instead of calculating all the qubit measurements
- Add measure qubit targets in the Circuit class
- Flip `requiresAllQubitsMeasurement` to False for the state vector and density matrix local simulators
- Update `simulator` to return a subset of measurements based on the measurement targets
- Add tests for the updated interpreter, simulator, and Circuit classes

*Testing done:*
tox

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/openqasm/_helpers/arrays.py
+++ b/src/braket/default_simulator/openqasm/_helpers/arrays.py
@@ -42,8 +42,6 @@ def convert_range_def_to_slice(range_def: RangeDefinition) -> slice:
     """Convert AST node into Python slice object"""
     buffer = np.sign(range_def.step.value) if range_def.step is not None else 1
     start = range_def.start.value if range_def.start is not None else None
-    if isinstance(range_def.end, UnaryExpression):
-        raise ValueError("Cannot measure non qubit or register values.")
     stop = (
         range_def.end.value + buffer
         if not (range_def.end is None or range_def.end.value == -1)
--- a/src/braket/default_simulator/openqasm/_helpers/arrays.py
+++ b/src/braket/default_simulator/openqasm/_helpers/arrays.py
@@ -33,6 +33,7 @@ from ..parser.openqasm_ast import (
     IntType,
     RangeDefinition,
     UintType,
+    UnaryExpression,
 )
 from .casting import LiteralType, cast_to, convert_string_to_bool_array
 
@@ -41,6 +42,8 @@ def convert_range_def_to_slice(range_def: RangeDefinition) -> slice:
     """Convert AST node into Python slice object"""
     buffer = np.sign(range_def.step.value) if range_def.step is not None else 1
     start = range_def.start.value if range_def.start is not None else None
+    if isinstance(range_def.end, UnaryExpression):
+        raise ValueError("Cannot measure non qubit or register values.")
     stop = (
         range_def.end.value + buffer
         if not (range_def.end is None or range_def.end.value == -1)
--- a/src/braket/default_simulator/openqasm/_helpers/arrays.py
+++ b/src/braket/default_simulator/openqasm/_helpers/arrays.py
@@ -33,7 +33,6 @@ from ..parser.openqasm_ast import (
     IntType,
     RangeDefinition,
     UintType,
-    UnaryExpression,
 )
 from .casting import LiteralType, cast_to, convert_string_to_bool_array
 
--- a/src/braket/default_simulator/openqasm/circuit.py
+++ b/src/braket/default_simulator/openqasm/circuit.py
@@ -41,6 +41,7 @@ class Circuit:
         self.instructions = []
         self.results = []
         self.qubit_set = set()
+        self.measured_qubits = []
 
         if instructions:
             for instruction in instructions:
@@ -59,6 +60,12 @@ class Circuit:
         """
         self.instructions.append(instruction)
         self.qubit_set |= set(instruction.targets)
+
+    def add_measure(self, target: int | list[int]):
+        if isinstance(target, list):
+            self.measured_qubits += target
+        else:
+            self.measured_qubits.append(target)
 
     def add_result(self, result: Results) -> None:
         """
--- a/src/braket/default_simulator/openqasm/circuit.py
+++ b/src/braket/default_simulator/openqasm/circuit.py
@@ -61,11 +61,11 @@ class Circuit:
         self.instructions.append(instruction)
         self.qubit_set |= set(instruction.targets)
 
-    def add_measure(self, target: int | list[int]):
-        if isinstance(target, list):
-            self.measured_qubits += target
-        else:
-            self.measured_qubits.append(target)
+    def add_measure(self, target: tuple[int]):
+        for qubit in target:
+            if qubit in self.measured_qubits:
+                raise ValueError(f"Qubit {qubit} is already measured or captured.")
+        self.measured_qubits += [qubit for qubit in target]
 
     def add_result(self, result: Results) -> None:
         """
--- a/src/braket/default_simulator/openqasm/circuit.py
+++ b/src/braket/default_simulator/openqasm/circuit.py
@@ -65,7 +65,7 @@ class Circuit:
         for qubit in target:
             if qubit in self.measured_qubits:
                 raise ValueError(f"Qubit {qubit} is already measured or captured.")
-        self.measured_qubits += [qubit for qubit in target]
+            self.measured_qubits.append(qubit)
 
     def add_result(self, result: Results) -> None:
         """
--- a/src/braket/default_simulator/openqasm/interpreter.py
+++ b/src/braket/default_simulator/openqasm/interpreter.py
@@ -470,7 +470,7 @@ class Interpreter:
 
     @visit.register
     def _(self, node: QuantumMeasurement) -> None:
-        qubits = self.context.get_qubits(node.qubit)
+        qubits = self.context.get_qubits(self.visit(node.qubit))
         self.context.add_measure(qubits)
 
     @visit.register
--- a/src/braket/default_simulator/openqasm/interpreter.py
+++ b/src/braket/default_simulator/openqasm/interpreter.py
@@ -470,15 +470,12 @@ class Interpreter:
 
     @visit.register
     def _(self, node: QuantumMeasurement) -> None:
-        qubit = node.qubit
-        if isinstance(qubit, IndexedIdentifier):
-            qubit_value = qubit.indices[0][0].value
-            self.context.add_measure(qubit_value)
-        else:  # Identifier
-            self.context.add_measure()
+        qubits = self.context.get_qubits(node.qubit)
+        self.context.add_measure(qubits)
 
     @visit.register
     def _(self, node: QuantumMeasurementStatement) -> None:
+        """The measure is performed but the assignment is ignored"""
         self.visit(node.measure)
 
     @visit.register
--- a/src/braket/default_simulator/openqasm/interpreter.py
+++ b/src/braket/default_simulator/openqasm/interpreter.py
@@ -470,11 +470,16 @@ class Interpreter:
 
     @visit.register
     def _(self, node: QuantumMeasurement) -> None:
-        """Doesn't do anything, but may add more functionality in the future"""
+        qubit = node.qubit
+        if isinstance(qubit, IndexedIdentifier):
+            qubit_value = qubit.indices[0][0].value
+            self.context.add_measure(qubit_value)
+        else:  # Identifier
+            self.context.add_measure()
 
     @visit.register
     def _(self, node: QuantumMeasurementStatement) -> None:
-        """Doesn't do anything, but may add more functionality in the future"""
+        self.visit(node.measure)
 
     @visit.register
     def _(self, node: ClassicalAssignment) -> None:
--- a/src/braket/default_simulator/openqasm/program_context.py
+++ b/src/braket/default_simulator/openqasm/program_context.py
@@ -890,3 +890,9 @@ class ProgramContext(AbstractProgramContext):
 
     def add_result(self, result: Results) -> None:
         self._circuit.add_result(result)
+
+    def add_measure(self, target: int = None):
+        if target is None:
+            self._circuit.add_measure(list(range(self.num_qubits)))
+        else:
+            self._circuit.add_measure(target)
--- a/src/braket/default_simulator/openqasm/program_context.py
+++ b/src/braket/default_simulator/openqasm/program_context.py
@@ -830,6 +830,9 @@ class AbstractProgramContext(ABC):
         """
         raise NotImplementedError
 
+    def add_measure(self, target: tuple[int]):
+        """Add qubit targets to be measured"""
+
 
 class ProgramContext(AbstractProgramContext):
     def __init__(self, circuit: Optional[Circuit] = None):
@@ -891,8 +894,5 @@ class ProgramContext(AbstractProgramContext):
     def add_result(self, result: Results) -> None:
         self._circuit.add_result(result)
 
-    def add_measure(self, target: int = None):
-        if target is None:
-            self._circuit.add_measure(list(range(self.num_qubits)))
-        else:
-            self._circuit.add_measure(target)
+    def add_measure(self, target: tuple[int]):
+        self._circuit.add_measure(target)
--- a/src/braket/default_simulator/openqasm/program_context.py
+++ b/src/braket/default_simulator/openqasm/program_context.py
@@ -119,12 +119,22 @@ class QubitTable(Table):
                 raise IndexError("Cannot index multiple dimensions for qubits.")
             primary_index = primary_index[0]
         if isinstance(primary_index, IntegerLiteral):
+            if primary_index.value >= len(self[name]):
+                raise IndexError(
+                    f"qubit register index `{primary_index.value}` out of range for qubit register `{name}`."
+                )
             target = (self[name][primary_index.value],)
         elif isinstance(primary_index, RangeDefinition):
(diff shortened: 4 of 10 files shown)


src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -60,6 +61,12 @@ def add_instruction(self, instruction: [GateOperation, KrausOperation]) -> None:
         self.instructions.append(instruction)
         self.qubit_set |= set(instruction.targets)
 
+    def add_measure(self, target: int | list[int]):
+        if isinstance(target, list):
+            self.measured_qubits += target
+        else:
+            self.measured_qubits.append(target)
```
ajberdy, line 68: for now, a set might be sufficient here. Eventually a set won't be enough information for mid-circuit measurements so this might be better

src/braket/default_simulator/openqasm/interpreter.py
```diff
@@ -470,11 +470,16 @@ def _(self, node: QuantumGateModifier) -> QuantumGateModifier:
 
     @visit.register
     def _(self, node: QuantumMeasurement) -> None:
-        """Doesn't do anything, but may add more functionality in the future"""
+        qubit = node.qubit
+        if isinstance(qubit, IndexedIdentifier):
+            qubit_value = qubit.indices[0][0].value
+            self.context.add_measure(qubit_value)
+        else:  # Identifier
+            self.context.add_measure()
```
ajberdy, line 478: looks like this case is missing an arg. I believe both cases can be simplified to just one using the helper `self.context.get_qubits(node.qubit)`, which takes an identifier (possibly indexed)

src/braket/default_simulator/openqasm/interpreter.py
```diff
@@ -470,11 +470,16 @@ def _(self, node: QuantumGateModifier) -> QuantumGateModifier:
...
 
     @visit.register
     def _(self, node: QuantumMeasurementStatement) -> None:
-        """Doesn't do anything, but may add more functionality in the future"""
+        self.visit(node.measure)
```
ajberdy, line 482: this isn't full functionality, as the assignment part of the statement will be ignored. That's alright for now, but I think we should update the comment/docstring to say that the measure is performed, but the assignment is ignored. Maybe even raising a warning

ajberdy: Though probably not raising a warning, since the BDK creates IR with measure assignments in the openqasm

src/braket/default_simulator/openqasm/interpreter.py
ajberdy: ^`get_qubits` will return a tuple of qubit indices, and can be used even with physical qubits (`$5`)

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -890,3 +890,9 @@ def add_kraus_instruction(self, matrices: list[np.ndarray], target: list[int]):
 
     def add_result(self, result: Results) -> None:
         self._circuit.add_result(result)
+
+    def add_measure(self, target: int = None):
+        if target is None:
+            self._circuit.add_measure(list(range(self.num_qubits)))
```
ajberdy, line 896: As above, I'm not sure when this case comes into play. Maybe this method should take a tuple though, since a measurement can be more than one qubit?

test/unit_tests/braket/default_simulator/openqasm/test_interpreter.py
```diff
@@ -1997,3 +1997,23 @@ def test_basis_rotation_hermitian():
...
+    circuit = Interpreter().build_circuit(qasm)
+    assert circuit.measured_qubits == [0]
+
+    qasm = """
+    bit[3] b;
+    qubit[3] q;
+    b = measure q;
+    """
+    circuit = Interpreter().build_circuit(qasm)
+    assert circuit.measured_qubits == [0, 1, 2]
```
ajberdy, line 2019: Can we make sure we're also testing:
- `measure q[0:1];`
- multiple measure statements (with and without overlap)
- multiple qubits/qubit registers defined
- `measure q[{0, 2}];`
- - `measure q[::-1];`
- remeasuring a qubit (should probably raise an error or be a noop, not sure yet)

ajberdy:
The failing dependent tests are pointing out a real issue. `Circuit.from_ir` uses the interpreter with a subclass of `ProgramContext`. Now when a `from_ir` call runs into a measurement, instead of performing a noop, it will try to use the `ProgramContext` method `add_measure`, which doesn't exist in the BDK `BraketProgramContext` subclass.

We'll need to add the method to the base class (likely as a noop)

src/braket/default_simulator/openqasm/interpreter.py
ashlhans: That is a great point, there is no reason to have add_measure not take an arg. I'll update add_measure to take in a required tuple and use `self.context.get_qubits(node.qubit)` here.

src/braket/default_simulator/openqasm/interpreter.py
ashlhans: I added this comment `"""The measure is performed but the assignment is ignored"""`. Do you think it would be worth adding the assignment functionality here as well?

src/braket/default_simulator/openqasm/program_context.py
ashlhans: Great point, I updated it to take a tuple instead.

src/braket/default_simulator/openqasm/_helpers/arrays.py
```diff
@@ -41,6 +42,8 @@ def convert_range_def_to_slice(range_def: RangeDefinition) -> slice:
     """Convert AST node into Python slice object"""
     buffer = np.sign(range_def.step.value) if range_def.step is not None else 1
     start = range_def.start.value if range_def.start is not None else None
+    if isinstance(range_def.end, UnaryExpression):
+        raise ValueError("Cannot measure non qubit or register values.")
```
ajberdy, line 46: This error seems out of place. I'm imagining you were motivated to add this because you got an error in the measurement visitor function. Make sure you're calling `self.visit` on the index before accessing it. I'll leave more detailed comments on that part of the code

src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -60,6 +61,12 @@ def add_instruction(self, instruction: [GateOperation, KrausOperation]) -> None:
         self.instructions.append(instruction)
         self.qubit_set |= set(instruction.targets)
 
+    def add_measure(self, target: tuple[int]):
+        for qubit in target:
+            if qubit in self.measured_qubits:
+                raise ValueError(f"Qubit {qubit} is already measured or captured.")
+        self.measured_qubits += [qubit for qubit in target]
```
ajberdy, line 68: Nit: this would be slightly more performant if you used a dict, which is guaranteed to maintain order of keys since python 3.6. Number of qubits is pretty small, so I think this is alright for now, but also wouldn't mind updating. Your call.

If you choose to keep as is, I'd leave a comment that it can be improved in the future.

src/braket/default_simulator/openqasm/interpreter.py
ajberdy: No, we can't actually do the assignment in this implementation (there is a feature branch that has assignment used by autoqasm). Here, all the OpenQASM interpreter logic happens as a first pass to create the "circuit", and then that is passed to the simulation. Only after the simulation do we get any measurements, so we would be well past this part of the code

src/braket/default_simulator/simulator.py
```diff
@@ -264,8 +265,12 @@ def _create_results_obj(
                 action=openqasm_ir,
             ),
             resultTypes=results,
-            measurements=self._formatted_measurements(simulation),
-            measuredQubits=self._get_measured_qubits(simulation.qubit_count),
+            measurements=self._formatted_measurements(simulation, measured_qubits),
+            measuredQubits=(
+                measured_qubits
+                if measured_qubits
+                else self._get_measured_qubits(simulation.qubit_count)
```
ajberdy, line 272: this method clearly doesn't fulfill its name anymore. Fortunately, it's a private method, so we can refactor the code here

src/braket/default_simulator/simulator.py
```diff
@@ -393,10 +400,15 @@ def _formatted_measurements(simulation: Simulation) -> list[list[str]]:
             list[list[str]]: List containing the measurements, where each measurement consists
             of a list of measured values of qubits.
         """
-        return [
+        # Get the full measurements
+        measurements = [
             list("{number:0{width}b}".format(number=sample, width=simulation.qubit_count))
             for sample in simulation.retrieve_samples()
         ]
+        #  Gets the subset or measurements from the full measurements
```
ajberdy, line 408: or -> of?

ashlhans: Yep, that should be `of`. Good catch!

src/braket/default_simulator/simulator.py
```diff
@@ -383,7 +388,9 @@ def _observable_hash(observable: Observable) -> Union[str, dict[int, str]]:
             return str(observable.__class__.__name__)
 
     @staticmethod
-    def _formatted_measurements(simulation: Simulation) -> list[list[str]]:
+    def _formatted_measurements(
+        simulation: Simulation, measured_qubits: list[int] = None
```
ajberdy, line 392:
```suggestion
        simulation: Simulation, measured_qubits: list[int] | None = None
```
and in the doc string

test/unit_tests/braket/default_simulator/openqasm/test_interpreter.py
```diff
@@ -1997,3 +1997,124 @@ def test_basis_rotation_hermitian():
...
+            [0, 2],
+        ),
+        (
+            "\n".join(
+                [
+                    "bit[1] b;",
+                    "qubit[2] q;",
+                    "h $0;",
+                    "cnot $0, $1;",
+                    "b[0] = measure $0;",
```
ajberdy, line 2074: Generally, a program can only have physical or virtual qubits (though I'm not sure that's enforced), so this test case shouldn't have `qubit[2] q;`

src/braket/default_simulator/openqasm/interpreter.py
```diff
@@ -470,11 +470,13 @@ def _(self, node: QuantumGateModifier) -> QuantumGateModifier:
 
     @visit.register
     def _(self, node: QuantumMeasurement) -> None:
-        """Doesn't do anything, but may add more functionality in the future"""
+        qubits = self.context.get_qubits(node.qubit)
```
ajberdy, line 473: you should visit `node.qubits` before calling get_qubits. For example if you have `q[i]` where is an int in scope, you will need to resolve that first. Try using `self.visit(node.qubits)` and that should resolve the issue you were having with range. I would also add a test for variables in a qubit index. Maybe
```
qubit[5] q;
for int i in [0:2] {
    measure q[i];
}
```
and assert measured qubits is [0, 1, 2]

src/braket/default_simulator/openqasm/circuit.py
ashlhans: Just to clarify, do you mean use a dict for `self.measured_qubits` here?

src/braket/default_simulator/simulator.py
```diff
@@ -393,10 +400,15 @@ def _formatted_measurements(simulation: Simulation) -> list[list[str]]:
...
         """
-        return [
+        # Get the full measurements
+        measurements = [
             list("{number:0{width}b}".format(number=sample, width=simulation.qubit_count))
             for sample in simulation.retrieve_samples()
         ]
+        #  Gets the subset or measurements from the full measurements
+        if measured_qubits is not None and measured_qubits != []:
+            measurements = [[row[i] for i in measured_qubits] for row in measurements]
```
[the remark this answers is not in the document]
ashlhans: I'm not sure that this would work since `sample` would be an integer so it cannot be indexed. 

For example, `simulation.retrieve_samples()` return something like:
`[7 0 7 1 7 0  1 0  0  1]`

and after the formatting returns the measurement list as 0s and 1s:
`[['1', '1', '0'], 
   ['1', '1', '1'], 
   ['0', '0', '1'], 
   ['0', '0', '1'], 
   ['1', '1', '0'], 
   ['1', '1', '0'], 
   ['1', '1', '1'], 
   ['1', '1', '1'], 
   ['0', '0', '1'], 
   ['1', '1', '1']]`

src/braket/default_simulator/openqasm/circuit.py
ajberdy: Per our offline discussion, we'll leave it as a list, since the number of qubits is small and in the future we may want to have duplicates

src/braket/default_simulator/simulator.py
ashlhans: I renamed this method to `_get_all_qubits` as that seems more accurate.

src/braket/default_simulator/openqasm/interpreter.py
ashlhans: Added `self.visit(node.qubits)` before getting the qubits and added a test.

src/braket/default_simulator/simulator.py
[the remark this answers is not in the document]
ajberdy: Oh fair enough. We could probably still simplify the nested list comprehension by using np indexing, but it's not the most important

test/unit_tests/braket/default_simulator/openqasm/test_interpreter.py
ashlhans: I will also add tests to validate the error message when an invalid qubit is measured (not in the range of qubits on the circuit or not a register)

src/braket/default_simulator/openqasm/circuit.py
ajberdy: We could append the qubits to measured_qubits during the loop to avoid iterating through target twice

ashlhans: Great idea, I'll move that inside the loop

src/braket/default_simulator/simulator.py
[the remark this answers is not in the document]
ashlhans: Updated to use np indexing like so: `measurements = np.array(measurements)[:, measured_qubits].tolist()`

test/unit_tests/braket/default_simulator/openqasm/test_interpreter.py
```diff
@@ -1997,3 +1997,124 @@ def test_basis_rotation_hermitian():
...
+        ),
+        (
+            "\n".join(
+                [
+                    "bit[1] b;",
+                    "qubit[3] q;",
+                    "h q[0];",
+                    "h q[1];",
+                    "cnot q[1], q[2];",
+                    "b[0] = measure q[::-1];",
```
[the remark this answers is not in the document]
ashlhans: Per our offline discussion, I'll avoid testing reversed indices in this PR.

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -119,12 +119,22 @@ def _(self, identifier: IndexedIdentifier) -> tuple[int]:
...
             target = tuple(np.array(self[name])[convert_range_def_to_slice(primary_index)])
         # Discrete set
         else:
-            target = tuple(np.array(self[name])[convert_discrete_set_to_list(primary_index)])
+            indices = convert_discrete_set_to_list(primary_index)
+            for index in indices:
+                if index >= len(self[name]):
+                    raise IndexError(
+                        f"qubit register index `{index}` out of range for qubit register `{name}`."
+                    )
```
ajberdy, line 136: maybe can we factor out common code between here and the integer case for error handling?

src/braket/default_simulator/simulator.py
```diff
@@ -383,20 +387,28 @@ def _observable_hash(observable: Observable) -> Union[str, dict[int, str]]:
...
         """
-        return [
+        # Get the full measurements
+        measurements = [
             list("{number:0{width}b}".format(number=sample, width=simulation.qubit_count))
             for sample in simulation.retrieve_samples()
         ]
+        #  Gets the subset of measurements from the full measurements
+        if measured_qubits is not None and measured_qubits != []:
+            measurements = np.array(measurements)[:, measured_qubits].tolist()
```
ajberdy, line 410: nice

test/unit_tests/braket/default_simulator/openqasm/test_interpreter.py
```diff
@@ -1997,3 +1997,199 @@ def test_basis_rotation_hermitian():
...
+    [
+        (
+            "\n".join(
+                [
+                    "bit[1] b;",
+                    "qubit[2] q;",
+                    "b[0] = measure q[5];",
+                ]
+            ),
+            "qubit register index `5` out of range for qubit register `q`.",
```
ajberdy, line 2179: maybe we should edit to "for qubit register of length 2 `q`"

src/braket/default_simulator/openqasm/program_context.py
ashlhans: Added an inner function with the common code.

test/unit_tests/braket/default_simulator/openqasm/test_interpreter.py
ashlhans: Updated the message.

ajberdy (approved):

PR #236, opened 2024-04-04
fix: support measurements on qubits without gates

ashlhans:
*Issue #, if available:*

*Description of changes:*
Currently, the local simulators do not support measuring qubits that are not used (do not a gate) on the circuit. 

Example:

`Circuit().measure(0)`

`Circuit().h(0).cnot(0,1).measure([0, 1, 2])`

To support this any qubits that are measured but not used, should return 0

Example:
```
Circuit().measure(0)

[[0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]
 [0]]

```
```
Circuit().h(0).cnot(0,1).measure([0, 1, 2])

[[1 1 0]
 [1 1 0]
 [0 0 0]
 [1 1 0]
 [1 1 0]
 [1 1 0]
 [0 0 0]
 [0 0 0]
 [1 1 0]
 [0 0 0]]
```

*Testing done:*
`tox` 

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/simulator.py
+++ b/src/braket/default_simulator/simulator.py
@@ -275,7 +275,7 @@ class BaseLocalSimulator(OpenQASMSimulator):
     @staticmethod
     def _validate_operation_qubits(operations: list[Operation]) -> None:
         qubits_referenced = {target for operation in operations for target in operation.targets}
-        if max(qubits_referenced) >= len(qubits_referenced):
+        if qubits_referenced and max(qubits_referenced) >= len(qubits_referenced):
             raise ValueError(
                 "Non-contiguous qubit indices supplied; "
                 "qubit indices in a circuit must be contiguous."
@@ -407,7 +407,37 @@ class BaseLocalSimulator(OpenQASMSimulator):
         ]
         #  Gets the subset of measurements from the full measurements
         if measured_qubits is not None and measured_qubits != []:
-            measurements = np.array(measurements)[:, measured_qubits].tolist()
+            if all(qubit in range(simulation.qubit_count) for qubit in measured_qubits):
+                measurements = np.array(measurements)[:, measured_qubits].tolist()
+            elif any(qubit in range(simulation.qubit_count) for qubit in measured_qubits):
+                measured_qubits_in_circuit = [
+                    qubit for qubit in measured_qubits if qubit in range(simulation.qubit_count)
+                ]
+                measured_qubits_not_in_circuit = [
+                    qubit for qubit in measured_qubits if qubit not in range(simulation.qubit_count)
+                ]
+
+                # get the list of measurements of qubits with gates on the circuit
+                measurements_in_circuit = np.array(measurements)[
+                    :, measured_qubits_in_circuit
+                ].tolist()
+                # set any measurements of qubits without gates to 0
+                measurements = np.zeros(
+                    (simulation.shots, len(measured_qubits_not_in_circuit)), dtype=int
+                ).tolist()
+
+                # Add measurements of qubits with gates on the circuit
+                for idx in measured_qubits_in_circuit:
+                    measurements[idx] = measurements_in_circuit[idx] + measurements[idx]
+
+                # Add measurements of 0s for qubits with no gates on the circuit
+                for idx, sublist in enumerate(measurements):
+                    if idx not in measured_qubits_in_circuit:
+                        measurements[idx] = measurements_in_circuit[idx] + measurements[idx]
+            else:
+                measurements = np.zeros(
+                    (simulation.shots, len(measured_qubits)), dtype=int
+                ).tolist()
         return measurements
 
     def run_openqasm(
--- a/test/unit_tests/braket/default_simulator/test_density_matrix_simulator.py
+++ b/test/unit_tests/braket/default_simulator/test_density_matrix_simulator.py
@@ -795,3 +795,42 @@ def test_measure_targets():
     assert 400 < np.sum(measurements, axis=0)[0] < 600
     assert len(measurements[0]) == 1
     assert result.measuredQubits == [0]
+
+
+def test_measure_no_gates():
+    qasm = """
+    bit[4] b;
+    qubit[4] q;
+    b[0] = measure q[0];
+    b[1] = measure q[1];
+    b[2] = measure q[2];
+    b[3] = measure q[3];
+    """
+    simulator = DensityMatrixSimulator()
+    result = simulator.run(OpenQASMProgram(source=qasm), shots=1000)
+    measurements = np.array(result.measurements, dtype=int)
+    assert np.sum(measurements, axis=0)[0] == 0
+    assert len(measurements[0]) == 4
+    assert result.measuredQubits == [0, 1, 2, 3]
+
+
+def test_measure_with_qubits_not_used():
+    qasm = """
+    bit[4] b;
+    qubit[4] q;
+    h q[0];
+    cnot q[0], q[1];
+    b[0] = measure q[0];
+    b[1] = measure q[1];
+    b[2] = measure q[2];
+    b[3] = measure q[3];
+    """
+    simulator = DensityMatrixSimulator()
+    result = simulator.run(OpenQASMProgram(source=qasm), shots=1000)
+    measurements = np.array(result.measurements, dtype=int)
+    assert 400 < np.sum(measurements, axis=0)[0] < 600
+    assert 400 < np.sum(measurements, axis=0)[1] < 600
+    assert np.sum(measurements, axis=0)[2] == 0
+    assert np.sum(measurements, axis=0)[3] == 0
+    assert len(measurements[0]) == 4
+    assert result.measuredQubits == [0, 1, 2, 3]
--- a/test/unit_tests/braket/default_simulator/test_state_vector_simulator.py
+++ b/test/unit_tests/braket/default_simulator/test_state_vector_simulator.py
@@ -1285,3 +1285,42 @@ def test_measure_targets():
     assert 400 < np.sum(measurements, axis=0)[0] < 600
     assert len(measurements[0]) == 1
     assert result.measuredQubits == [0]
+
+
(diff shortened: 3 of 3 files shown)


test/unit_tests/braket/default_simulator/test_density_matrix_simulator.py
```diff
@@ -795,3 +795,42 @@ def test_measure_targets():
...
+
+
+def test_measure_no_gates():
+    qasm = """
+    bit[4] b;
+    qubit[4] q;
+    b[0] = measure q[0];
+    b[1] = measure q[1];
+    b[2] = measure q[2];
+    b[3] = measure q[3];
```
mbeach-aws, line 807: since we know the number of qubits (i.e. we have `qubit[4] q;`), can we do `b = measure q;` ? I assume the 4 is because there's gates/measurements on 4 qubits, so the BDK does know that information.

mbeach-aws (changes requested):

test/unit_tests/braket/default_simulator/test_density_matrix_simulator.py
ashlhans: We can do `b = measure q`. I can update this test

ashlhans: The BDK would be able to know the number of qubits on the circuit based on the gates/measurements added. 
The measurement targets have to be specified though.

src/braket/default_simulator/simulator.py
```diff
@@ -407,7 +407,37 @@ def _formatted_measurements(
...
+                    measurements[idx] = measurements_in_circuit[idx] + measurements[idx]
+
+                # Add measurements of 0s for qubits with no gates on the circuit
+                for idx, sublist in enumerate(measurements):
+                    if idx not in measured_qubits_in_circuit:
+                        measurements[idx] = measurements_in_circuit[idx] + measurements[idx]
+            else:
+                measurements = np.zeros(
+                    (simulation.shots, len(measured_qubits)), dtype=int
+                ).tolist()
```
speller26, line 440: Do these need to be separate cases? From what I can tell, the second branch covers everything

ashlhans: Yeah, you're right. The first and last branch are not needed. In fact, there shouldn't need to be a check if the qubits are in the simulation. I'll simplify this down

src/braket/default_simulator/simulator.py
```diff
@@ -407,7 +407,28 @@ def _formatted_measurements(
...
+            # get the list of measurements of qubits with gates on the circuit
+            measurements_in_circuit = np.array(measurements)[:, measured_qubits_in_circuit].tolist()
+            # set any measurements of qubits without gates to 0
+            measurements = np.zeros(
+                (simulation.shots, len(measured_qubits_not_in_circuit)), dtype=int
+            ).tolist()
+
+            # Add measurements of qubits with gates on the circuit
+            for idx in measured_qubits_in_circuit:
+                measurements[idx] = measurements_in_circuit[idx] + measurements[idx]
```
speller26, line 426: Not sure I understand what's going on here; `idx` is the qubit index, but the item of `measurements` you're changing is indexed by the shots. I'm even more confused by the fact that it works.

test/unit_tests/braket/default_simulator/test_state_vector_simulator.py
```diff
@@ -1285,3 +1285,39 @@ def test_measure_targets():
...
+    b[0] = measure q[0];
+    b[1] = measure q[1];
+    b[2] = measure q[2];
+    b[3] = measure q[3];
+    """
+    simulator = StateVectorSimulator()
+    result = simulator.run(OpenQASMProgram(source=qasm), shots=1000)
+    measurements = np.array(result.measurements, dtype=int)
+    assert np.sum(measurements, axis=0)[2] == 0
+    assert len(measurements[0]) == 4
```
speller26, line 1303: Ditto

src/braket/default_simulator/simulator.py
```diff
@@ -407,7 +407,28 @@ def _formatted_measurements(
...
+
+            # Add measurements of 0s for qubits with no gates on the circuit
+            for idx, sublist in enumerate(measurements):
+                if idx not in measured_qubits_in_circuit:
+                    measurements[idx] = measurements_in_circuit[idx] + measurements[idx]
```
speller26, line 431: Why not just use [`np.pad`](https://numpy.org/doc/stable/reference/generated/numpy.pad.html)?
```python
measurements = np.pad(measurements, ((0, 0), (0, len(measured_qubits_not_in_circuit)))).tolist()
```

src/braket/default_simulator/simulator.py
```diff
@@ -407,7 +407,28 @@ def _formatted_measurements(
...
+            measurements = np.zeros(
+                (simulation.shots, len(measured_qubits_not_in_circuit)), dtype=int
+            ).tolist()
+
+            # Add measurements of qubits with gates on the circuit
+            for idx in measured_qubits_in_circuit:
+                measurements[idx] = measurements_in_circuit[idx] + measurements[idx]
+
+            # Add measurements of 0s for qubits with no gates on the circuit
+            for idx, sublist in enumerate(measurements):
```
speller26, line 429: No need to `enumerate`; the value `sublist` isn't used anywhere

src/braket/default_simulator/simulator.py
ashlhans: `np.pad` will work so long as the circuit is not empty. If the circuit is empty `Circuit().measure(0)` the initial measurements is already a list of 0s so there is one extra row. I could fix this by handling the empty circuit in another if statement branch.

src/braket/default_simulator/simulator.py
ashlhans: Good point. I updated this to use the your suggested `np.pad()` so this is all removed.

test/unit_tests/braket/default_simulator/test_density_matrix_simulator.py
```diff
@@ -795,3 +795,39 @@ def test_measure_targets():
...
+    """
+    simulator = DensityMatrixSimulator()
+    result = simulator.run(OpenQASMProgram(source=qasm), shots=1000)
+    measurements = np.array(result.measurements, dtype=int)
+    assert np.sum(measurements, axis=0)[0] == 0
+    assert len(measurements[0]) == 4
```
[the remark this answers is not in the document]
ashlhans: Good point, updated this.

test/unit_tests/braket/default_simulator/test_state_vector_simulator.py
ashlhans: Good point, updated this.

src/braket/default_simulator/simulator.py
ashlhans: Switched to using the padding solution instead.

src/braket/default_simulator/simulator.py
speller26: Why would `Circuit().measure(0)` create a list of zeros rather than a list of empty lists (within the default simulator implementation)?

speller26: Got it, it's from
```
"{number:0{width}b}".format(number=sample, width=simulation.qubit_count)
```

speller26 (approved):

src/braket/default_simulator/simulator.py
```diff
@@ -407,7 +407,21 @@ def _formatted_measurements(
         ]
         #  Gets the subset of measurements from the full measurements
         if measured_qubits is not None and measured_qubits != []:
-            measurements = np.array(measurements)[:, measured_qubits].tolist()
+            if any(qubit in range(simulation.qubit_count) for qubit in measured_qubits):
+                measured_qubits_in_circuit = [
```
mbeach-aws, line 411: this part does two list-comprehensions to find qubits greater or less than simulation.qubit_count. 

This could be more efficient using numpy with boolean indexing, which is vectorized and very fast. Additionally, the condition is only written once below:
```
measured_qubits = np.array(measured_qubits)
in_circuit_mask = measured_qubits < simulation.qubit_count

measured_qubits_in_circuit = measured_qubits[in_circuit_mask]

measured_qubits_not_in_circuit = measured_qubits[~in_circuit_mask]
```

src/braket/default_simulator/simulator.py
```diff
@@ -407,7 +407,21 @@ def _formatted_measurements(
...
+                    qubit for qubit in measured_qubits if qubit < simulation.qubit_count
+                ]
+                measured_qubits_not_in_circuit = [
+                    qubit for qubit in measured_qubits if qubit >= simulation.qubit_count
+                ]
+                measurements = np.array(measurements)[:, measured_qubits_in_circuit].tolist()
+                measurements = np.pad(
+                    measurements, ((0, 0), (0, len(measured_qubits_not_in_circuit)))
+                ).tolist()
```
mbeach-aws, line 420: I find the conversions between lists and numpy pretty confusing. 

Is there a way to keep them as numpy arrays? Maybe somethign like this:

```    
measurements_array = np.array(measurements)
selected_measurements = measurements_array[:, measured_qubits_in_circuit]


measurements = np.pad(selected_measurements, ((0, 0), (0, len(measured_qubits_not_in_circuit)))).tolist()
```

mbeach-aws (changes requested):

src/braket/default_simulator/simulator.py
ashlhans: Updated this to use boolean indexing.

src/braket/default_simulator/simulator.py
ashlhans: Updated to keep `measurements_array` and `selected_measurements` as a numpy array.

mbeach-aws (approved):

speller26 (approved):

PR #260, opened 2024-06-04
feat: Introduce MultiSimulator class

speller26:
*Issue #, if available:*

*Description of changes:*

Introduces MultiSimulator class to allow backends to leverage their own batching implementations. Will next publish SDK PR to make use of this interface.

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/simulator/braket_simulator.py
+++ b/src/braket/simulator/braket_simulator.py
@@ -63,3 +63,33 @@ class BraketSimulator(ABC):
     @abstractmethod
     def properties(self) -> DeviceCapabilities:
         """DeviceCapabilities: Properties of the device."""
+
+
+class MultiSimulator(BraketSimulator):
+    """An abstract simulator that can run multiple tasks locally.
+
+    The purpose of this interface is to take advantage of any circuit batching (or more generally,
+    multithreading/multiprocessing) that the backend can leverage to improve performance.
+
+    The tasks can be either quantum circuits defined as OpenQASM or JAQCD programs,
+    or analog Hamiltonian simulation (AHS) programs.
+    """
+
+    @abstractmethod
+    def run_multiple(
+        self, payloads: list[Union[OQ3Program, AHSProgram, JaqcdProgram]], *args, **kwargs
+    ) -> list[Union[GateModelTaskResult, AnalogHamiltonianSimulationTaskResult]]:
+        """
+        Run the tasks specified by the given IR payloads.
+
+        Extra arguments will contain any additional information necessary to run the tasks,
+        such as number of qubits.
+
+        Args:
+            payloads (list[Union[OQ3Program, AHSProgram, JaqcdProgram]]): The IR representations
+                of the programs
+
+        Returns:
+            list[Union[GateModelTaskResult, AnalogHamiltonianSimulationTaskResult]]: A list of
+            result objects, with the ith object being the result of the ith program.
+        """
--- a/src/braket/default_simulator/simulator.py
+++ b/src/braket/default_simulator/simulator.py
@@ -13,7 +13,7 @@
 
 import uuid
 import warnings
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from collections.abc import Callable
 from typing import Any, Union
 
@@ -61,7 +61,7 @@ _NOISE_INSTRUCTIONS = frozenset(
 )
 
 
-class OpenQASMSimulator(BraketSimulator, ABC):
+class OpenQASMSimulator(BraketSimulator):
     """An abstract simulator that runs an OpenQASM 3 program.
 
     Translation of individual operations and observables from OpenQASM to the desired format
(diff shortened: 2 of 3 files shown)


kshyatt-aws (dismissed):

src/braket/simulator/braket_simulator.py
```diff
@@ -63,3 +63,33 @@ def run(
     @abstractmethod
     def properties(self) -> DeviceCapabilities:
         """DeviceCapabilities: Properties of the device."""
+
+
+class MultiSimulator(BraketSimulator):
```
rmshaffer, line 68: can we also export this `MultiSimulator` from `__init__.py` in this module? to allow
`from braket.simulator import MultiSimulator`
instead of having to do
`from braket.simulator.braket_simulator import MultiSimulator`

krneta, line 68: It seems like a little overkill to create a new abstract class/interface this one method that's very related to the existing run method.

src/braket/simulator/braket_simulator.py
```diff
@@ -63,3 +63,33 @@ def run(
...
+
+    The purpose of this interface is to take advantage of any circuit batching (or more generally,
+    multithreading/multiprocessing) that the backend can leverage to improve performance.
+
+    The tasks can be either quantum circuits defined as OpenQASM or JAQCD programs,
+    or analog Hamiltonian simulation (AHS) programs.
+    """
+
+    @abstractmethod
+    def run_multiple(
```
krneta, line 79: We should probably have a default implementation of this, so that future simulators don't need to re-implement it.

src/braket/simulator/braket_simulator.py
rmshaffer: Yes, I'm also curious why we couldn't just add `run_multiple` to the existing interface, and provide a default implementation.

src/braket/simulator/braket_simulator.py
rmshaffer: Why do we call it `run_multiple` instead of `run_batch`? It's also a little confusing because it's unspecified how the *args and **kwargs should apply to the batch. For example, do they apply globally to all of the tasks? Or should these be lists of the (*args, **kwargs) that would be provided to a single `run` call?

src/braket/simulator/braket_simulator.py
```diff
@@ -63,3 +63,33 @@ def run(
...
+        self, payloads: list[Union[OQ3Program, AHSProgram, JaqcdProgram]], *args, **kwargs
+    ) -> list[Union[GateModelTaskResult, AnalogHamiltonianSimulationTaskResult]]:
+        """
+        Run the tasks specified by the given IR payloads.
+
+        Extra arguments will contain any additional information necessary to run the tasks,
+        such as number of qubits.
```
rmshaffer, line 86: I know this is copied from the `run` docstring, but "number of qubits" is a weird example here. Better examples would be number of shots, or lists of input parameters to the tasks.

src/braket/simulator/braket_simulator.py
speller26: Regardless, we're going to need a new method for the backend to run multiple circuits; this means it has to live either in BraketSimulator or a new class. If it lives in `BraketSimulator`, then the SDK would need either check whether the method throws an exception (not great), or we'd need to add a new property in `BraketSimulator` for the SDK to check, which then leads to a situation where devs can implement the method and forget to mark the property true, or vice-versa.

src/braket/simulator/braket_simulator.py
speller26: The SDK already provides the default implementation if the simulator doesn't subclass `MultiSimulator`; if someone doesn't want a native `run_multiple` method, they can just implement `BraketSimulator`.

As for `run_multiple` vs `run_batch`, the reasons are twofold:
1. The simulator doesn't really have any notion of "batch" the way the SDK does; the method just returns a list of results.
2. The word "batch" is already used in the state vector simulator for subdividing the circuit into chunks of gates and contracting these "batches" together (I'm sorry about the name, my bad), so I didn't want to reuse it.

src/braket/simulator/braket_simulator.py
rmshaffer: Could `BraketSimulator` also provide a default implementation of `run_multiple`? (maybe it could do exactly what the SDK does today for `run_batch` - delegate to `Pool`)

speller26: I suppose it could; it just doesn't seem useful to have the same implementation in both the simulator _and_ the SDK; as well, the abstract `BraketSimulator` class having a default implementation could potentially lead to multiple inheritance issues down the line.

speller26: Alternative implementation in #264

rmshaffer:
@speller26 should we close this now in favor of https://github.com/amazon-braket/amazon-braket-default-simulator-python/pull/264?

PR #283, opened 2025-05-27
infra: performance optimization for the default sim using ping-pong buffers, gate fusion, and optimized gate paths

AbeCoull:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests


AbeCoull:
Coverage doesn't like `numba` decorated functions. Working on a fix.

setup.py
```diff
@@ -31,6 +31,7 @@
     package_data={"": ["*.g4", "*.inc"]},
     include_package_data=True,
     install_requires=[
+        "numba",
```
rmshaffer, line 34: Adding the `numba` dependency is probably the riskiest component of this change. I'm not sure how much of a concern this is. `numba` pulls in `llvmlite`, which is not a small package, but I don't know of any specific compatibility problems.

One option to reduce the risk: we could move this dependency into an extras package. So a user would have to explicitly do something like `pip install amazon-braket-default-simulator[numba]`. Then, you could dynamically check if `numba` is available, e.g. [as done here](https://stackoverflow.com/a/71247818/13124175).

We could default to including it in our containers and notebooks, but this would at least provide a way for users to install the simulator without `numba` (in case of some conflict, or in case they need to minimize package downloads). But again, this may be unnecessary right now, and would be easy to add later if we find a specific scenario where it's required.

AbeCoull: https://numba.pydata.org/numba-doc/0.42.0/user/installing.html#compatibility - There is a fairly high level of support across all platforms.

I shall explore that option. Presumably if a customer is running circuits large enough to need numba, then they are already in a situation where they have a powerful enough piece of hardware to process through. I am not partial to further conditional branching based on package inclusion as that opens up a few pitfalls so deferring to when this becomes an issue/customer signal may be an option here. 

This optional dependency can also be linked to extra package data where these functions live to avoid install checks but the dispatch logic will be slightly more complicated there

rmshaffer: > deferring to when this becomes an issue/customer signal may be an option here

I agree with this approach!

src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
```diff
@@ -641,25 +641,16 @@ def _apply_hamiltonian(
             "is used as an approximation."
         )
 
-    # If the integrator uses intermediate time value that is larger than the minimum
-    # time value specified, the final time value is used as an approximation.
-    if index_time < 0:
```
speller26, line 646: Why remove?

src/braket/analog_hamiltonian_simulator/rydberg/scipy_solver.py
```diff
@@ -105,27 +107,27 @@ def f(index_time: int, y: np.ndarray) -> scipy.sparse.csr_matrix:
...
             )
 
-        integrator.set_initial_value(state, index_time)
         integrator.integrate(index_time + 1)
 
-        # get the current state, and normalize it
         state = integrator.y
         state /= np.linalg.norm(state)  # normalize the state
-        states.append(state)
+        states[index_time + 1] = state
```
speller26, line 128: Ah, the changes for AHS deserve a separate PR

src/braket/analog_hamiltonian_simulator/rydberg/rydberg_simulator_helpers.py
AbeCoull: We have a time check in the caller function. I can revert the AHS stuff for a separate PR

src/braket/analog_hamiltonian_simulator/rydberg/scipy_solver.py
AbeCoull: Ack, will work on the revert

test/unit_tests/braket/default_simulator/test_density_matrix_simulation.py
```diff
@@ -223,3 +329,133 @@ def test_simulation_retrieve_samples():
...
+        (
+            "Multi-qubit fusion",
+            2,
+            [gate_operations.CX([0, 1]), gate_operations.CX([0, 1])],
+            [],
+            1000,
+        ),
+    ],
+)
+def test_gate_fusion_equivalence(test_case, qubit_count, operations1, operations2, samples):
```
speller26, line 433: Is this test relevant for this PR?

PR #361, opened 2026-04-15
fix: add a circuit validation test and github action

AbeCoull:
*Issue #, if available:*

*Description of changes:*
adds in a circuit validation suite to github actions and tox

*Testing done:*

## Merge Checklist

#### General

#### Tests


tox.ini
```diff
@@ -10,6 +10,14 @@ commands =
     pytest {posargs} --cov-report term-missing --cov-report html --cov-report xml --cov=braket
 extras = test
 
+[testenv:circuits-validation]
```
rmshaffer: I suggest calling this `integ-tests` and putting the new file under `test/integ_tests`, for consistency with how the SDK repo does it: https://github.com/amazon-braket/amazon-braket-sdk-python/blob/85bdb9e1d806f1d4723b1e116f8507ee61488aa7/tox.ini#L24-L36

.github/workflows/circuits-validation.yml
```diff
@@ -0,0 +1,34 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+      - feature/**
+
+permissions:
+  contents: read
```
rmshaffer: since the triggers are the same - any reason not to just add this new job to https://github.com/amazon-braket/amazon-braket-default-simulator-python/blob/main/.github/workflows/build.yml rather than making a new workflow file?

tox.ini
AbeCoull: Sure, I shall update it. It falls a bit closer to a correctness testing workflow which is why I left it with such a specific env name but this is a 2 way door which we can change later if there is a more fitting env name such as `correctness_testing`.

.github/workflows/circuits-validation.yml
AbeCoull: I missed we made that change to actions - adding this in

rmshaffer (approved):

test/integ_tests/circuits_test.py
```diff
@@ -0,0 +1,886 @@
...
+def _to_braket(ops, n_qubits):
+    c = Circuit()
+    for gate_name, targets in ops:
+        getattr(c, gate_name)(*targets)
+    for q in range(n_qubits):
+        c.measure(q)
+    return c
+
+
+def _measured_probabilities(ops, n_qubits):
```
aniksd-braket, line 81: Why can't we use `result.measurement_probabilities` in this function, and then filter out based on `n_qubits`

aniksd-braket (commented):
This is a great extension to our testing capability. My one concern would be the time it would now take to run our CI tests, do you have any insights/metrics into that?

aniksd-braket (approved):

rmshaffer:
> This is a great extension to our testing capability. My one concern would be the time it would now take to run our CI tests, do you have any insights/metrics into that?

They're actually run in this PR itself. Looks like the new integ-tests took only about 1 minute.

PR #366, opened 2026-04-28
fix: use tracking-based is_mcm_dependent in AbstractProgramContext

rmshaffer:
*Issue #, if available:*
  
*Description of changes:*
  
`AbstractProgramContext.is_mcm_dependent` now uses the scope-tracking machinery that already exists on the base class (`_mcm_dependent_scopes`, `_is_name_mcm_dependent`, `_referenced_identifiers`). The identical override on `ProgramContext` is removed.
  
Subclasses that use a different tracking model (e.g.`qiskit-braket-provider`'s `_QiskitProgramContext`, which tracks measured bits in its own set) should override `is_mcm_dependent` to reflect that model. A companion change in `qiskit-braket-provider` does exactly this.
  
  *Testing done:*
  
  - All existing unit tests pass.
  - Updated `test_is_mcm_dependent_defaults_to_*` to reflect the new semantics (no measurements tracked → not MCM-dependent).
  - Verified the pending `qiskit-braket-provider` changes in https://github.com/amazon-braket/qiskit-braket-provider/pull/323 pass on their own against the released default simulator, as well as against this branch.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/openqasm/program_context.py
+++ b/src/braket/default_simulator/openqasm/program_context.py
@@ -917,18 +917,25 @@ class AbstractProgramContext(ABC):
     def is_mcm_dependent(self, expression) -> bool:
         """Whether an expression depends on any mid-circuit measurement result.
 
-        An expression is MCM-dependent when any identifier it references
-        resolves (via lexical scoping) to a variable that was produced by
-        a mid-circuit measurement. Subclasses populate
-        ``_mcm_dependent_scopes`` (typically via ``track_mcm_dependency``
-        or when a measurement is recorded) so this check walks each
-        referenced identifier's scope stack and stops at the scope where
-        the name is declared.
+        The default returns ``True`` so that any subclass overriding
+        ``supports_midcircuit_measurement = True`` routes control flow
+        and classical operations through its MCM-aware handlers
+        (``evaluate_condition``, ``evaluate_for_range``,
+        ``evaluate_while_condition``, ``iter_classical_scopes``). This
+        matches the pre-tracking semantics where any MCM-capable context
+        handled every occurrence.
 
-        Used by the Interpreter to decide whether control flow and
-        classical assignments need per-path evaluation. Expressions that
-        are not MCM-dependent are evaluated once and eagerly, matching
-        non-MCM behavior.
+        Subclasses that implement MCM dependency tracking (by populating
+        ``_mcm_dependent_scopes`` via ``track_mcm_dependency`` and on
+        measurement) should override this to return a real check, e.g.::
+
+            return any(
+                self._is_name_mcm_dependent(name)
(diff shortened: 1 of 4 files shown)


src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -919,28 +919,33 @@ def add_verbatim_marker(self, marker) -> None:
...
         Args:
             expression: The AST expression to check.
 
         Returns:
             bool: True if the expression depends on an MCM result.
         """
-        return any(
-            self._is_name_mcm_dependent(name) for name in self._referenced_identifiers(expression)
-        )
+        return True
```
speller26: Not sure that subclasses should be responsible for determining whether an expression depends on an MCM; that should be implementation-independent. The previous documentation was incorrect in that the determination is in fact being done at the abstract level, but the problem is expressions that use MCM results but have not yet been fed to a condition.

speller26:
Implemented a different fix in #367

src/braket/default_simulator/openqasm/program_context.py
yitchen-tim: I think the root cause is that the tracking `is_mcm_dependent` only happens in ProgramContext class, not in the AbstractProgramContext. So the downstream, program context like the Qiskit one miss this implementation. We should move this logic [here](https://github.com/amazon-braket/amazon-braket-default-simulator-python/blob/3efd89cfe8099ee61e25c8de6a8e8d7fc8ff547d/src/braket/default_simulator/openqasm/program_context.py#L1514) to AbstractProgramContext (or equivalently, implement that in QiskitProgramContext)

rmshaffer: Yes, the only reason I had this return `True` here is that it seems like the safest default for a class (such as this abstract base class, or anything that inherits from its defaults) that does no tracking of MCMs at all. However, I defer to you both on whether this is the right place to do it. (Although - this does fix the downstream tests, so it could work even if just for the short-term.)

rmshaffer: Iterating on this to try out some things, removing the `return True`.

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -1092,7 +1092,7 @@ def iter_classical_scopes(self, expression):
         Yields:
             None: Signals the Interpreter to evaluate the expression once.
         """
-        raise NotImplementedError
+        yield
```
speller26, line 1095: I actually originally had this default, but decided against it because I don't want the abstract class to bias towards an implementation (in this case translation)

rmshaffer: Yeah, makes sense. Happy to close this PR in favor of #367

PR #374, opened 2026-05-05
fix: Rename `Measure` to `Projection`

speller26:
This has caused me a bit of confusion while running some tests, so I'm renaming `Measure` to something more accurate.

*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/gate_operations.py
+++ b/src/braket/default_simulator/gate_operations.py
@@ -1284,7 +1284,7 @@ class GPhase(GateOperation):
 class Projection(GateOperation):
     """
     Post-selected projector onto a single qubit's ``|outcome⟩`` subspace,
-    with renormalization.
+    with normalization.
 
     Used by branched simulation: when a mid-circuit measurement forks a
     path, each branch records its known outcome and replays the path's
--- a/src/braket/default_simulator/gate_operations.py
+++ b/src/braket/default_simulator/gate_operations.py
@@ -1281,64 +1281,51 @@ class GPhase(GateOperation):
         return "gphase"
 
 
-class Measure(GateOperation):
+class Projection(GateOperation):
     """
-    Measurement operation that projects the state to a specific outcome.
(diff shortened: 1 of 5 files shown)


rmshaffer (commented):
**edit:** I'm realizing now my comments below don't make much sense because this PR actually eliminates the case where no outcome is specified. So, this is not really a concern once I realized that.

--

If I understand your motivation correctly, you're trying to address the confusion that comes from the `Measure` operation optionally having a pre-specified outcome, which is more accurately described as a `Projection`.

However, when there isn't a pre-specified outcome, wouldn't `Measure` still be the more intuitive name?

I'm wondering if perhaps we should actually factor out the pre-specified `result` capability from `Measure`, and move that to a new `Projection` class which (possibly, if it reduces duplication) subclasses `Measure`.

speller26:
Yeah, through multiple refactors, the `Measure` operation has had responsibilities successively removed until it no longer did anything other than project onto an outcome.

src/braket/default_simulator/gate_operations.py
```diff
@@ -1281,64 +1281,51 @@ def gate_type(self) -> str:
...
 
-class Measure(GateOperation):
+class Projection(GateOperation):
     """
-    Measurement operation that projects the state to a specific outcome.
-
-    This is used in branched simulation to apply measurement projections
-    when recalculating states from instruction sequences.
+    Post-selected projector onto a single qubit's ``|outcome⟩`` subspace,
+    with renormalization.
```
yitchen-tim:
```suggestion
    with normalization.
```

yitchen-tim (approved):

src/braket/default_simulator/gate_operations.py
speller26: Heh, it would be pretty confusing to see the word "renormalization" here

PR #389, opened 2026-07-14
fix: inner-scope shadowing during branched identifier lookup

### LINKED ISSUE #386: LocalSimulator crash: 'FloatLiteral' object has no attribute 'name' when a classical variable is present with a measurement-conditioned if (default-simulator 1.39.5)
### Summary
On the latest `amazon-braket-default-simulator` (1.39.5), a **valid** OpenQASM 3 program crashes with an
internal `AttributeError: 'FloatLiteral' object has no attribute 'name'` when it combines a classical
scalar variable with a measurement-dependent `if`. The program parses cleanly (see *Validity* below); the
crash happens during **execution**.

### Reproduction (crashes)
```python
from braket.devices import LocalSimulator
from braket.ir.openqasm import Program

qasm = """
OPENQASM 3;
include "stdgates.inc";
qubit[1] q;
bit c;
float a = 3.14159265;
c = measure q[0];
if (c == false) { }
rx(a) q[0];
"""
LocalSimulator().run(Program(source=qasm), shots=10).result()
```
```
AttributeError: 'FloatLiteral' object has no attribute 'name'
```

### It runs fine without the classical variable
Dropping `float a` and using a literal angle -- `rx(3.14159265) q[0];` -- runs fine, as does removing the
`if`. The crash needs **both** a classical scalar variable **and** a measurement-conditioned branch. It is
not specific to angles: `int a = 3; c = measure q[0]; if (c == false) { } x q[0];` crashes the same way
(`'IntegerLiteral' object has no attribute 'name'`) -- an *unused* scalar + a measurement branch + any
following gate is enough. (`angle` gives `'SymbolLiteral' object has no attribute 'name'`.)

### Validity (this is a valid program, not a syntax error)
The program parses cleanly with **both** the reference parser and Braket's own parser, so this is not a
malformed-input rejection:
```python
import openqasm3
openqasm3.parse(qasm)                                    # OK: 7-statement Program AST, round-trips

from braket.default_simulator.openqasm import interpreter
interpreter.parse(qasm)                                  # OK: Program AST, 7 statements
```
The `AttributeError` is raised later, during **execution** (the interpreter visit), not at parse time.

### Traceback (key frames)
```
.../braket/default_simulator/openqasm/interpreter.py", line 452, in _
    [self.context.qubit_mapping.get_qubit_size(qubit) for qubit in qubits]
.../braket/default_simulator/openqasm/program_context.py", line 184, in get_qubit_size
    return len(self.get_by_identifier(identifier))
.../braket/default_simulator/openqasm/program_context.py", line 140, in get_by_identifier
    if identifier.name.startswith("$"):
AttributeError: 'FloatLiteral' object has no attribute 'name'
```
On the measurement-branched execution path, the gate's qubit resolution appears to receive the classical
value (a `FloatLiteral` / `IntegerLiteral` / `SymbolLiteral`) where a qubit `Identifier` is expected.

### Expected
The program should simulate. A classical variable together with a measurement-conditioned `if` is valid
OpenQASM 3, so it should not raise an internal `AttributeError`.

### Environment
- `amazon-braket-default-simulator` **1.39.5** (latest on PyPI, 2026-06-22)
- `amazon-braket-sdk` 1.121.0 - `amazon-braket-schemas` 1.31.0 - `openqasm3` 1.0.1 - `antlr4-python3-runtime` 4.13.2
- Python 3.12.9 - Windows 11 - device: `LocalSimulator()` (StateVectorSimulator)

rmshaffer:
*Issue #, if available:*

Fixes #386.

*Description of changes:*
The branched-mode overrides of `get_value_by_identifier` and `is_initialized` consult the per-path variable table before the scope-aware `variable_table`. When a user-defined gate expansion declares a qubit alias (e.g. `rx`'s parameter `a`) whose name matches an outer-scope classical variable, that outer variable lives in the flat per-path dict and shadows the inner qubit alias, crashing on the next `get_qubit_size` call with `'FloatLiteral' object has no attribute 'name'`.

Defer to the superclass path whenever the identifier is bound in a non-global `variable_table` scope so inner-scope bindings win, matching the non-branched execution semantics.

*Testing done:*

tox passes. Test added which reproduced existing bug, now passes.

## Merge Checklist

#### General

#### Tests

### DIFF
--- a/src/braket/default_simulator/openqasm/program_context.py
+++ b/src/braket/default_simulator/openqasm/program_context.py
@@ -1258,6 +1258,9 @@ class ProgramContext(AbstractProgramContext):
             return super().get_value_by_identifier(identifier)
 
         name = get_identifier_name(identifier)
+        if self._bound_in_inner_scope(name):
+            return super().get_value_by_identifier(identifier)
+
         path = self._paths[self._active_path_indices[0]]
         framed_var = path.get_variable(name)
         if framed_var is None:
@@ -1285,6 +1288,9 @@ class ProgramContext(AbstractProgramContext):
         if not self._is_branched:
             return super().is_initialized(name)
 
+        if self._bound_in_inner_scope(name):
+            return super().is_initialized(name)
+
         # Check per-path variables first
         path = self._paths[self._active_path_indices[0]]
         framed_var = path.get_variable(name)
@@ -1293,6 +1299,10 @@ class ProgramContext(AbstractProgramContext):
 
         # Fall back to shared variable table
         return super().is_initialized(name)
+
+    def _bound_in_inner_scope(self, name: str) -> bool:
+        """Whether ``name`` is bound in a non-global variable-table scope."""
+        return any(name in scope for scope in self.variable_table._scopes[1:])
 
     def _flush_pending_mcm_for_variable(self, name: str) -> None:
         """If ``name`` matches a pending MCM's classical destination, flush it.
--- a/src/braket/default_simulator/openqasm/program_context.py
+++ b/src/braket/default_simulator/openqasm/program_context.py
@@ -1258,14 +1258,12 @@ class ProgramContext(AbstractProgramContext):
             return super().get_value_by_identifier(identifier)
 
         name = get_identifier_name(identifier)
-        if self._bound_in_inner_scope(name):
-            return super().get_value_by_identifier(identifier)
-
         path = self._paths[self._active_path_indices[0]]
         framed_var = path.get_variable(name)
-        if framed_var is None:
-            # Fall back to the shared variable table for variables declared
-            # before branching started
+        # Fall back to the shared variable table for variables declared before
+        # branching started and for identifiers shadowed by an inner scope
+        # (e.g. gate-expansion qubit aliases).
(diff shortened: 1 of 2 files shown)


AJ-comp:
I'm the reporter of #386. I checked out this branch and ran the original repro from the issue — it passes. Thanks for tracking it down.

speller26:
> When a user-defined gate expansion declares a qubit alias (e.g. rx's parameter a) whose name matches an outer-scope classical variable, that outer variable lives in the flat per-path dict and shadows the inner qubit alias, crashing on the next get_qubit_size call with 'FloatLiteral' object has no attribute 'name'.

Can you clarify what you mean by declaring `a` as a qubit alias?

rmshaffer:
> Can you clarify what you mean by declaring a as a qubit alias?

Maybe "alias" is the wrong word, I just mean when it is used as a qubit argument in a `gate` definition, e.g.:
```
gate rx(theta) a {
  U(theta, -1.57, 1.57) a;
}
```
The problem comes when the program then has another type of variable, e.g. a `float`, also named `a`.

speller26 (approved):

PR #399, opened 2026-08-10
feat: Output variable implementation

speller26:
*Issue #, if available:*

*Description of changes:*

*Testing done:*

## Merge Checklist

#### General

#### Tests


src/braket/default_simulator/openqasm/circuit.py
```diff
@@ -67,13 +67,28 @@ def add_measure(
...
+                Default: None.
+            allow_remeasure (bool): Whether to allow remeasuring a qubit. Default: False.
+
+        Returns:
+            list[int]: The classical indices assigned for each measured qubit.
+        """
+        assigned_indices = []
         for index, qubit in enumerate(target):
             classical_index = (
                 classical_targets[index]
```
yitchen-tim, line 87: This looks wrong. You only open fresh slot when classical_targets is None. For cases like 
```
c1 = measure q1;     
c2[0] = measure q2;
```
will silently be incorrect

src/braket/default_simulator/openqasm/output_helpers.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    output_variables = context.output_variables
+    bindings = context.output_bindings
+    classical_idx_columns = _column_map(context)
+    outputs = []
+    shot_offset = 0
+    for path in context.active_paths:
+        base_values = {name: _path_base_value(context, path, name) for name in output_variables}
+        for shot_idx in range(shot_offset, shot_offset + path.shots):
+            row = measurements[shot_idx] if shot_idx < len(measurements) else None
```
yitchen-tim: When does the else branch happen? Should we actually throw error instead of padding None here?

test/unit_tests/braket/default_simulator/openqasm/test_program_context.py
```diff
@@ -173,3 +177,77 @@ def test_add_barrier_is_noop():
...
+def test_output_binding_indexed_destination():
+    context = run_output_program(
+        """
+        OPENQASM 3.0;
+        output bit[2] c;
+        qubit[2] q;
+        c[1] = measure q[0];
+        """
+    )
+    assert context.output_bindings == {"c": {1: 1}}
```
yitchen-tim, line 234: shouldn't it start with 0? Like `{"c": {1: 0}}`

test/unit_tests/braket/default_simulator/openqasm/test_program_context.py
```diff
@@ -173,3 +177,77 @@ def test_add_barrier_is_noop():
...
+    context = run_output_program(
+        """
+        OPENQASM 3.0;
+        output int x;
+        bit[2] c;
+        qubit[2] q;
+        c = measure q;
+        """
+    )
+    assert context.output_bindings == {}
```
yitchen-tim, line 247: Why there is no binding here? Is it because we expect output int x; throw error in the upstream?

src/braket/default_simulator/openqasm/program_context.py
```diff
@@ -2008,11 +2075,12 @@ def _branch_measurement(
         self._measure_and_branch(target)
         self._update_classical_from_measurement(target, classical_destination)
         if classical_targets is not None:
-            self._circuit.add_measure(
+            assigned_indices = self._circuit.add_measure(
                 target,
                 classical_targets,
                 allow_remeasure=self.supports_midcircuit_measurement,
             )
+            self._record_output_binding(classical_destination, assigned_indices)
```
yitchen-tim: We need to treat the case like `output bit[2] c;`, right? Currently this statement skips this code path

test/unit_tests/braket/default_simulator/openqasm/test_program_context.py
```diff
@@ -15,11 +15,15 @@
```
yitchen-tim, line 15: Now that we are supporting multiple registers, we need a lot more extensive tests here to make sure the indexing is correct. For example, we need
- mix usage of indexed and non-indexed declarations
- mix usage of indexed and non-indexed measure assignments
- using multiple register with MCM and control flow to make sure the single path method and branch method both work

We also want end-to-end tests

src/braket/default_simulator/openqasm/output_helpers.py
speller26: It shouldn't happen, but this is a public function that can technically take unrelated data. I was on the fence about throwing an error, and happy to change the implementation to do that instead.

test/unit_tests/braket/default_simulator/openqasm/test_program_context.py
speller26: This is just an artifact of the way the simulator fills classical variable slots; since we asked for `c[1]`, it assigns slot 1 of the tracked variables.

test/unit_tests/braket/default_simulator/openqasm/test_program_context.py
speller26: `output_bindings` is used only to track output registers.

src/braket/default_simulator/openqasm/program_context.py
speller26: Can you clarify what you mean here? This is only called when the measurement is performed.

test/unit_tests/braket/default_simulator/openqasm/test_program_context.py
speller26: I've been tinkering with property-based testing, but I want to introduce that in a later PR

